PR #25, opened 2017-07-19
fscrypt PAM module

### LINKED ISSUE #4: Add PAM module
As mentioned in the `README`, the idea of protecting directories with a login passphrase is incomplete without a [PAM module](http://www.linux-pam.org/Linux-PAM-html). This PAM module will need to support the following functionality:

- Automatically re-wrapping a user's login protector when their PAM passphrase changes.
- Automatically unlocking directories on login

The road map for changes to incorporate the PAM module is as follows:

- [x] (#21) Refactoring the `pam` package to support the additional functionality. 
- [x] (#3, #18) Improvements to the `actions` package making unlocking Policies easier. See #3 and #18 
- [x] (#25) Writing the PAM module itself. This will be a package exporting the correct identifiers built as a `.so`.
- [x] (#25) Adding build support in the Makefile
- [x] (#25) Updating the Documentation

josephlr:
This commit closes #4 by adding in `pam_fscrypt`. See the updated documentation for more information.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,4 @@
-/fscrypt
+fscrypt
+fscrypt.*
 fscrypt_image
+pam_fscrypt.so
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,9 +30,10 @@ these commands when writing your code.
 
 ### Building and Testing
 
-As mentioned in `README.md`, running `make` will build the fscrypt executable.
-Running `make go` will build each package and run the tests, but just running
-`make go` with nothing else will skip the integration tests.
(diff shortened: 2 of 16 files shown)


fancytenseletters:
There is https://github.com/neithernut/pam_e4crypt available. You can look at the code if there something interesting code to borrow.

josephlr:
Thanks for bringing this up! We definitely saw this, but couldn't actually use any of the code (due to GPLv2 vs Apache license compatibility).  There is also the issue that `e4crypt` is basically going away in favor of a simpler tool like [fscryptctl](https://github.com/google/fscryptctl) that is not just aimed toward one filesystem.

PR #39, opened 2017-08-18
Purge command now clears cache

### LINKED ISSUE #35: Automatically clear cache when fscrypt purge is used
As described in https://github.com/google/fscrypt#i-can-still-see-files-or-filenames-after-running-fscrypt-purge-mountpoint encryption keys are still available in kernel cache after issuing `fscrypt purge` command.

There is however option to clear kernel cache by changing  `/proc/sys/vm/drop_caches` value to `2` or `3`. That means ` fscrypt purge` could send something like `sync && echo 3 > /proc/sys/vm/drop_caches` and the cache will be cleared. The downside is that ` fscrypt purge` would need execution with `sudo`.

josephlr:
Fixes #35 

This PR also does some general cleanup by separating things like privilege dropping and keyring management to a separate `security` package. 

As part of this cleanup,  the `(Drop/Raise)ThreadPrivileges` functions are added. These also serve to make sure the correct keyrings are linked. These will also be used with the PAM module (see #4)

### DIFF
--- a/security/keyring.go
+++ b/security/keyring.go
@@ -95,7 +95,7 @@ var keyringIDCache = make(map[int]int)
 // simpler approach would be to use
 //     unix.KeyctlGetKeyringID(unix.KEY_SPEC_USER_KEYRING, false)
 // which would work in almost all cases. However, despite the fact that the rest
-// of the keyrings API using the _effective_ UID throughout, the translation of
+// of the keyrings API uses the _effective_ UID throughout, the translation of
 // KEY_SPEC_USER_KEYRING is done with respect to the _real_ UID. This means that
 // a simpler implementation would not respect permissions dropping.
 func getUserKeyringID() (int, error) {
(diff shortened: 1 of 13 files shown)


security/keyring.go
```diff
@@ -0,0 +1,159 @@
...
+	return nil
+}
+
+var keyringIDCache = make(map[int]int)
+
+// getUserKeyringID returns the key id of the current user's user keyring. A
+// simpler approach would be to use
+//     unix.KeyctlGetKeyringID(unix.KEY_SPEC_USER_KEYRING, false)
+// which would work in almost all cases. However, despite the fact that the rest
+// of the keyrings API using the _effective_ UID throughout, the translation of
```
josephlr: Typo using -> uses

security/keyring.go
```diff
@@ -0,0 +1,159 @@
...
+		keyringIDCache[euid] = keyringID
+		return keyringID, nil
+	}
+
+	return 0, ErrFindingKeyring
+}
+
+func keyringLink(keyID int, keyringID int) error {
+	_, err := unix.KeyctlInt(unix.KEYCTL_LINK, keyID, keyringID, 0, 0)
+	return errors.Wrapf(err, "linking key %d into keyring %d", keyID, keyringID)
```
josephlr: Add dedicated error type here so that we can switch on it later.

PR #70, opened 2017-10-05
Typos

kant:


josephlr:
Thanks for the fixes. #71 should hopefully prevent this in the future.

PR #84, opened 2018-02-09
Move to x/crypto/argon2

### LINKED ISSUE #82: Move to a pure go implementation of Argon2id
An implementation of Argon2id has [made it into x/crypto](https://go-review.googlesource.com/c/crypto/+/91935)!!!

We should switch over to this as it would get rid of any dependency on libargon2 and potentially increase performance. I'll post more perf info when I have a chance to prototype.

josephlr:
Fixes #82 

This change removes any dependency on [`libargon2`](https://github.com/P-H-C/phc-winner-argon2) by switching to [`x/crypto/argon2`](https://godoc.org/golang.org/x/crypto/argon2). This simplifies both the build-time and run-time dependencies for fscrypt. While the new implementation is slightly slower, it will have the chance to improve with [changes to the Go language/compiler](https://tip.golang.org/doc/go1.10#asm).

This change also vendors the needed dependencies.


josephlr:
CLA should be fixed

PR #98, opened 2018-03-26
[WIP] Move `fscrypt.conf`'s format to TOML

### LINKED ISSUE #80: Move `fscrypt.conf`'s format to TOML
`fscrypt.conf` is currently formatted as a JSON file. This is undesirable for many reasons.
- JSON is [not a great format for a configuration file](https://revelry.co/json-configuration-file-format/)
- JSON does not support comments (without a [dirty hack](https://stackoverflow.com/a/244858))
- JSON is designed to be human-readable *not* human-writable

During the initial design process for fscrypt, [we considered using TOML](https://docs.google.com/document/d/1GBbuCMe-wLuXosVI87cv8CuRkqwncpnNdXAJDECP-6w/edit#heading=h.fnje9u7ksk2l) as the config format (like [rust does](https://users.rust-lang.org/t/why-does-cargo-use-toml/3577)). The main reason we didn't is because TOML is less common than JSON and because the [main go library](https://github.com/BurntSushi/toml) had a [legally problematic license](https://github.com/BurntSushi/toml/issues/165).

Now that TOML is used by toolchain programs like [`dep`](https://github.com/golang/dep) and the [licensing problem is fixed](https://github.com/BurntSushi/toml/issues/183), TOML makes more sense.

The following will need to be completed to make this change:
- [ ] Add support for reading `fscrypt.conf` as TOML
- [ ] Start generating all new `fscrypt.conf` files as TOML
- [ ] Warn users who have JSON formatted `fscrypt.conf` files
- [ ] Remove support for reading a JSON config file (would need to be done before 1.0)

vbmade2000:
Fixes #80

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -36,6 +36,12 @@
   revision = "d6ce6262d87e3a4e153e86023ff56ae771554a41"
 
 [[projects]]
+  name = "github.com/pelletier/go-toml"
+  packages = ["."]
+  revision = "acdc4509485b587f5e675510c4f2c63e90ff68a8"
+  version = "v1.1.0"
+
+[[projects]]
   name = "github.com/pkg/errors"
   packages = ["."]
   revision = "645ef00459ed84a119197bfb8d8205042c6df63d"
@@ -114,6 +120,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "13bc740ba83539fecd6dd4b7d026fd221077369cb6b170e83d288383d0fde474"
+  inputs-digest = "c4fcfe791713d5df5b10ce69471f48baef680bdb5f6756ef41853d5956862838"
   solver-name = "gps-cdcl"
   solver-version = 1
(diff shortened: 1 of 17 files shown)


vbmade2000:
I signed it!

vbmade2000:
@josephlr  I tried following command sequence.

**1. fscrypt setup (Worked successfully)
2. fscrypt setup /tmp/testmount (Worked successfully)
3. fscrypt encrypt /tmp/testmount**

Third command raises following error.

> fscrypt encrypt: /tmp/testmount/: not an empty directory
> 
> Encryption can only be setup on empty directories; files cannot be encrypted in-place.
> Instead, encrypt an empty directory, copy the files into that encrypted directory, and
> securely delete the originals with "shred".

Both testmount and mydata dir contain only .`fscrypt `directory and nothing else.

josephlr:
@vbmade2000 sorry for not responding on this, it feel off my radar. 

I don't think changing the format to TOML at this point will work (given the number of users it might break). Thanks anyway for the work on this.

PR #114, opened 2018-12-04
Makefile: use a specific protoc-gen-go version

ebiggers:
'make gen' no longer works because it uses the git version of
protoc-gen-go, which is no longer compatible with the latest released
version of github.com/golang/protobuf/proto, which we're using.  Freeze
the protoc-gen-go version so that it keeps working.


josephlr:
Thanks for the fix Eric! I need to figure out a way to fix the versions of the tools via go modules.

PR #119, opened 2019-01-15
Add support for the Adiantum encryption mode

ebiggers:
Add Adiantum support to the fscrypt userspace tool.  Supported in the
kernel since v5.0-rc1, Adiantum is a length-preserving encryption mode
based primarily on XChaCha12.  It is fast even on CPUs without AES
instructions.  Unlike XTS it is also a wide-block encryption mode.
Adiantum is supported for both contents and filenames encryption.

For Adiantum encryption policies, also make the fscrypt tool provide the
new DIRECT_KEY flag, which further improves performance by requesting
that all files be encrypted directly with the policy key.  This takes
advantage of Adiantum's support for long tweaks.

See the kernel commit "fscrypt: add Adiantum support"
(https://git.kernel.org/torvalds/c/8094c3ceb21ad938) for more details.


ebiggers:
> @ebiggers is there any plan for other upstream ciphers to ever support the "direct key" configuration?

It will be supported for any other long-tweak modes that may happen to be added in the future, but so far Adiantum is the only one.  It doesn't work for XTS and CBC.   There's a chance we'll implement it for XTS and CBC by hashing longer tweaks down to 16 bytes, but there isn't as much of a reason to.

PR #142, opened 2019-08-25
Fix 'make format'-related CI failure

ebiggers:
* Exclude generated files from 'make format'
* Run 'make format' with latest version of goimports


ebiggers:
Updated this pull request to actually fix the CI failure... Turns out that `make gen` was also conflicting with `make format`.

PR #148, opened 2019-09-18
Filesystem keyring and v2 encryption policy support

ebiggers:
This pull request adds support for the filesystem keyring and v2 encryption policies, which are new features in the kernel part of fscrypt in Linux v5.4.  It also adds an `fscrypt lock` command, to mirror `fscrypt unlock`.

Users can opt into using v2 policies on new encrypted directories by setting `"policy_version": "2"` in `/etc/fscrypt.conf`.  These directories always use the filesystem keyring, and they can be unlocked and locked by non-root users.

Alternatively, users can opt into using the filesystem keyring for existing directories by setting `"use_fs_keyring_for_v1_policies": true`.  This keeps the directories compatible with older kernels, but makes unlocking and locking them start requiring root.

The filesystem keyring solves the issues users have been running into where processes unexpectedly don't get access to unlocked encrypted files.  It also makes locking directories faster, more reliable, and not always require root.

For more information, see the [cover letter of the kernel patchset](https://lkml.kernel.org/linux-fsdevel/20190805162521.90882-1-ebiggers@kernel.org/), the [updated kernel documentation](https://git.kernel.org/pub/scm/fs/fscrypt/fscrypt.git/tree/Documentation/filesystems/fscrypt.rst), and the [kernel pull request](https://lkml.kernel.org/lkml/20190916044258.GA8269@sol.localdomain/).  Similar changes are [being made to Android](https://android-review.googlesource.com/q/topic:fscrypt-key-mgmt-improvements+(status:open+OR+status:merged)) too.


josephlr:
Eric, this is great!! This will take me some time to get through, but thank you so much for splitting it up into individual commits.

A quick question. How much complexity is reduced if we don't bother with `use_fs_keyring_for_v1_policies`? We could either add this feature in later, or just make it easier for users to update their directories.

ebiggers:
> A quick question. How much complexity is reduced if we don't bother with `use_fs_keyring_for_v1_policies`? We could either add this feature in later, or just make it easier for users to update their directories.

The diff from what I have now would look something like this: https://github.com/ebiggers/fscrypt/commit/f05f15ae873245f32f1fb15b842a5e2c2a44a46c

It's not a huge difference since we have to support the filesystem keyring anyway.  But supporting `use_fs_keyring_for_v1_policies` does mean we have to document that setting, handle re-elevating to root to access the keyring in the PAM module, and a few other small things.

My main concern with *not* having this setting is that fscrypt will still be creating v1 encrypted directories by default for some time, and it's rather unfriendly to users to say "hey, you created the wrong type of encrypted directory, please delete everything and start over".

(I think I did screw it up slightly, though.  As I've currently written it, `"use_fs_keyring_for_v1_policies": true` will break fscrypt if you downgrade your kernel.  We should check that the ioctls are supported before we honor the setting.)

Maryse47:
@josephlr could you consider merging this one? Linux 5.4 is around the corner and it would be good to have userspace that has v2 encryption policy support.

ebiggers:
> (I think I did screw it up slightly, though. As I've currently written it, "use_fs_keyring_for_v1_policies": true will break fscrypt if you downgrade your kernel. We should check that the ioctls are supported before we honor the setting.)

I've fixed this now: `use_fs_keyring_for_v1_policies` is only honored if the kernel supports it.

Also, Linux v5.4 has been released now, and I've updated the sys/unix package upstream with the new fscrypt declarations.  I've rebased this onto https://github.com/google/fscrypt/pull/169 which does the sys/unix upgrade.

@josephlr can you review this when you have a chance?  Thanks!

josephlr:
> @josephlr can you review this when you have a chance? Thanks!

I’m currently in India, so I might not get to this until after Christmas, but will review as soon as I can.

ebiggers:
> > @josephlr can you review this when you have a chance? Thanks!
> 
> I’m currently in India, so I might not get to this until after Christmas, but will review as soon as I can.

Okay, in the mean time do you mind if I start merging preparatory commits and some of my other pull requests that should be non-controversial?  I'm thinking these commits:

* travis: set GO111MODULE=on for 'go get'
* Allow filesystem links to contain leading/trailing whitespace
* cmd/fscrypt: adjust message when listing protector sources
* Upgrade to latest golang.org/x/sys module
* Use latest fscrypt declarations from sys/unix
* Rename some variables from 'target' to 'targetUser'
* cmd/fscrypt: preserve paragraphs in wrapText()

josephlr:
@ebiggers that seems reasonable to me

ebiggers:
Rebased onto latest master branch, and updated to make `fscrypt status` on an incompletely locked directory show `Unlocked: Partially (incompletely locked)` rather than `Unlocked: No`, as suggested by @mhogomchungu at https://github.com/ebiggers/fscrypt/commit/e4de7b4a524eef89eaf4279a05306e094a3539f6#commitcomment-36229352.

ebiggers:
@josephlr can you please review this?

FYI, I'm also thinking of adding support for setting `policy_version: "auto"` in /etc/fscrypt.conf, which would make v2 policies be used iff the kernel supports them; and possibly making that the default setting.  That can be done later, though.

ebiggers:
This has been out for review for 4 months now, and people keep running into the problems this fixes.  So I'm going ahead and merging it.  @josephlr please feel free to still review it later, and I'd be glad to address any suggestions.

Issue #129, opened 2019-05-18
fscrypt checks wrong device for setup and throws ErrNotSetup


Redsandro:
My `/home` is a separate partition. 

```
$ mount | grep /dev/sd
/dev/sda5 on / type ext4 (rw,relatime,errors=remount-ro,data=ordered)
/dev/sda6 on /home type ext4 (rw,relatime,data=ordered)
```

I want to test out fscrypt on a test directory.

```
$ sudo fscrypt setup
$ sudo fscrypt setup /home
$ sudo mkdir -p /home/test/enc
$ sudo chown -R redsandro: /home/test
$ sudo fscrypt encrypt /home/test/enc --user=redsandro

Your data can be protected with one of the following sources:
1 - Your login passphrase (pam_passphrase)
2 - A custom passphrase (custom_passphrase)
3 - A raw 256-bit key (raw_key)
Enter the source number for the new protector [2 - custom_passphrase]: 1

fscrypt encrypt: filesystem /: not setup for use with fscrypt

Run "fscrypt setup MOUNTPOINT" to use fscrypt on this filesystem.
```

See: __fscrypt encrypt: filesystem /: not setup for use with fscrypt__

I believe that the state of `/` (AKA `/dev/sda5`) is irrelevant.
This is about `/home` (AKA `/dev/sda6`).

```
$ sudo tune2fs -l /dev/sda6 | grep encrypt
Filesystem features:
    has_journal
    ext_attr
    resize_inode
    dir_index
    filetype
    needs_recovery
    extent
    flex_bg
    encrypt        # <- here it is
    sparse_super
    large_file
    huge_file
    uninit_bg
    dir_nlink
    extra_isize
```

Redsandro:
This seems to work:

```
cd /home/test
sudo fscrypt encrypt enc --user=redsandro
```

I'm not sure if the path from where the command is executed is technically desirable, so I'll leave this issue open.

josephlr:
@Redsandro Thanks for reporting this issue. Your intuition is correct, you should not need to enable encryption on the `/` filesystem to encrypt a directory on the `/home` filesystem. However, when creating a "pam_passphrase" protector (i.e. a protector linked to your login passphrase), fscrypt stores some data on the root partition, specifically a wrapped key that lets your passphrase keep working even if your system password changes.

[...]

Issue #59, opened 2017-09-05
[question] Should bind-mounts be treated as separate filesystems?


ghost:
I noticed that currently fscrypt treats directories mounted on existing filesystem with` --bind` option as separate mountpoints.
That means it needs executing `fscrypt setup MOUNTPOINT` on every directory mounted this way to make encryption available. Moreover it can create confusion when encrypted directory was created before bind-mounts were used: 
`fscrypt setup /`
`fscrypt encrypt /home/xxx` (creates policy under `/.fscrypt`)
bind-mount /home to /home
`fscrypt setup /home`
`fscrypt status /home/xxx` (searches policy under `/home/.fscrypt`)

In that case user have to manually copy fscrypt metadata from  `/.fscrypt` to `/home/.fscrypt`

ebiggers:
In my opinion bind mounts shouldn't be treated as separate filesystems, since otherwise it's ambiguous where the fscrypt metadata is located.  I've opened a pull request to fix this: https://github.com/google/fscrypt/pull/154

PR #149, opened 2019-09-23
cmd/fscrypt: make 'fscrypt setup' create /.fscrypt

ebiggers:
Make the global setup command also create the metadata directory at
/.fscrypt, since that's where login protectors are placed, even when the
actual encrypted directories are on a different filesystem.

Resolves https://github.com/google/fscrypt/issues/129

### DIFF
--- a/cmd/fscrypt/commands.go
+++ b/cmd/fscrypt/commands.go
@@ -41,37 +41,49 @@ var Setup = cli.Command{
 	ArgsUsage: fmt.Sprintf("[%s]", mountpointArg),
 	Usage:     "perform global setup or filesystem setup",
 	Description: fmt.Sprintf(`This command creates fscrypt's global config
-		file or enables fscrypt on a filesystem.
+		file and/or prepares a filesystem for use with fscrypt.
 
-		(1) When used without %[1]s, create the parameters in %[2]s.
-		This is primarily used to configure the passphrase hashing
-		parameters to the appropriate hardness (as determined by %[3]s).
-		Being root is required to write the config file.
+		(1) When used without %[1]s, this command creates the global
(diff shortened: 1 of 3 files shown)


Maryse47:
Does it solve https://github.com/google/fscrypt/issues/59 as well?

ebiggers:
> Does it solve #59 as well?

No, bind mounts are a separate issue.

cmd/fscrypt/commands.go
```diff
@@ -63,6 +69,14 @@ func setupAction(c *cli.Context) error {
 	case 0:
 		// Case (1) - global setup
 		err = createGlobalConfig(c.App.Writer, actions.ConfigFileLocation)
+		if err != nil {
```
josephlr: This code here is essentially what we want to do, but it might make the code more readable to just change this all to something like:

```go
switch c.NArg() {
case 0:
	// Case (1) - global setup
	if err := createGlobalConfig(c.App.Writer, actions.ConfigFileLocation); err != nil {
		return newExitError(c, err)
	}
	if err := setupFilesystem(c.App.Writer, "/"); err != nil {
		if errors.Cause(err) != filesystem.ErrAlreadySetup {
			return newExitError(c, err)
		}
		fmt.Fprintf(c.App.Writer, "Skipping creating /.fscrypt because it already exists.\n")
	}
case 1:
	// Case (2) - filesystem setup
	if err := setupFilesystem(c.App.Writer, c.Args().Get(0)); err != nil {
		return newExitError(c, err)
	}
default:
	return expectedArgsErr(c, 1, true)
}
return nil
``` 

And avoid all the assigning to `err` stuff.

ebiggers: Agreed, updated now.

Issue #131, opened 2019-06-17
compatibility between ostree and fscrypt


ramcq:
I am experimenting with fscrypt integrated into Endless OS (endlessos.com), an ostree (https://ostree.readthedocs.io/en/latest/) managed Linux system. When booted by the ostree module in the initrd, a root filesystem is selected from `/ostree/deploy/yada...` and ends up being bind mounted as `/` (typically read-only) and the original filesystem `/` is copied to `/sysroot`.

In many cases ostree uses a symlink from `/` to `/var` or `/sysroot` for things like `/home` and `/root` which are expected to be mutable and persist across OS deployments, so my first idea was to add a `.fscrypt` symlink to `sysroot/.fscrypt` - which works provided I have invoked setup on the `/sysroot` path - however if I have not, and I attempt to call setup on `/` with a dangling symlink, it tries to set up in `/` and fails due to the path existing (as a dangling symlink).

So; two possible approaches - in the case of a dangling symlink, follow to the target and go there to create the dir - or detect ostree booted systems by checking for `/run/ostree-booted` and switching the metadata path for `/` to `/sysroot` automatically. Any thoughts/preferences?

ebiggers:
I've implemented the solution where `/.fscrypt` can be a symlink that points to a writable location: https://github.com/google/fscrypt/pull/150.  Does this work for you?

ramcq:
Reading through the code, this looks ideal. Thanks for your help on this and your work on fscrypt in general! Much appreciated. I'll add testing it to my todo list - understandably it's a little precarious to do so on my normal system.

ramcq:
I've tested in a VM, this works great. Thanks!

PR #150, opened 2019-09-25
filesystem: allow .fscrypt to be a symlink

ebiggers:
Support the case where the user has a read-only root filesystem (e.g.
with OSTree) and had previously created a symlink /.fscrypt pointing to
a writable location, so that login protectors can be created there.

Resolves https://github.com/google/fscrypt/issues/131

### DIFF
--- a/filesystem/filesystem.go
+++ b/filesystem/filesystem.go
@@ -85,6 +85,10 @@ var (
 // used when a Policy on filesystem A is protected with Protector on filesystem
 // B. In this scenario, we store a "link file" in the protectors directory whose
 // contents look like "UUID=3a6d9a76-47f0-4f13-81bf-3332fbe984fb".
+//
+// We also allow ".fscrypt" to be a symlink which was previously created. This
+// allows login protectors to be created when the root filesystem is read-only,
+// provided that "/.fscrypt" is a symlink pointing to a writable location.
 type Mount struct {
 	Path       string
 	Filesystem string
@@ -124,9 +128,21 @@ func (m *Mount) String() string {
 	Device:    %s`, m.Path, m.Filesystem, m.Options, m.Device)
 }
 
-// BaseDir returns the path of the base fscrypt directory on this filesystem.
+// BaseDir returns the path to the base fscrypt directory for this filesystem.
 func (m *Mount) BaseDir() string {
-	return filepath.Join(m.Path, baseDirName)
+	rawBaseDir := filepath.Join(m.Path, baseDirName)
+	// We allow the base directory to be a symlink, but some callers need
+	// the real path, so dereference the symlink here if needed. Since the
+	// directory the symlink points to may not exist yet, we have to read
+	// the symlink manually rather than use filepath.EvalSymlinks.
+	target, err := os.Readlink(rawBaseDir)
+	if err != nil {
+		return rawBaseDir // not a symlink
+	}
+	if filepath.IsAbs(target) {
+		return target
+	}
+	return filepath.Join(m.Path, target)
 }
 
 // ProtectorDir returns the directory containing the protector metadata.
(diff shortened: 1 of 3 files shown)


filesystem/filesystem.go
```diff
@@ -129,6 +133,23 @@ func (m *Mount) BaseDir() string {
 	return filepath.Join(m.Path, baseDirName)
 }
 
+// baseDirReal is like BaseDir, except it evaluates the symlink if the BaseDir
+// is a symlink.
+func (m *Mount) baseDirReal() (string, error) {
```
josephlr: So this approch will work, but it seems like it would be simpler to just have `BaseDir` call [`filepath.EvalSymlinks`](https://golang.org/pkg/path/filepath/#EvalSymlinks) and just return that for all `BaseDir` operations. That would also help keep the symlink code size down and in one place.

```go
// BaseDir returns the path of the base fscrypt directory on this filesystem after resolving symlinks.
func (m *Mount) BaseDir() string {
	dir, err := filepath.EvalSymlinks(filepath.Join(m.Path, baseDirName))
        if err != nil {
                panic(err)
        }
        return dir
}

File: filesystem/filesystem.go [hunk 2]
```diff
@@ -226,8 +252,13 @@ func (m *Mount) Setup() error {
...
 	}
 
+	dstPath, err := m.baseDirReal()
+	if err != nil {
+		return m.err(err)
+	}
+
 	// Atomically move directory into place.
-	return m.err(os.Rename(temp.BaseDir(), m.BaseDir()))
+	return m.err(os.Rename(temp.BaseDir(), dstPath))
```
josephlr: This `os.Rename` code here and below could then stay the same.

filesystem/path.go
```diff
@@ -56,6 +56,16 @@ func loggedStat(name string) (os.FileInfo, error) {
 	return info, err
 }
 
+// loggedLstat runs os.Lstat (doesn't dereference trailing symlink), but it logs
```
josephlr: This file would then be unnecessary.

filesystem/filesystem_test.go
```diff
@@ -109,6 +110,67 @@ func TestRemoveAllMetadata(t *testing.T) {
...
+		t.Fatal("real base dir should no longer exist")
+	}
+}
+
+func TestSetupWithAbsoluteSymlink(t *testing.T) {
+	mnt, err := getTestMount(t)
+	if err != nil {
+		t.Fatal(err)
+	}
+	realDir, err := ioutil.TempDir("", ".fscrypt-real")
```
josephlr: What's the reason for creating this directory just to remove it?

filesystem/filesystem.go
ebiggers: That doesn't work because `EvalSymlinks` returns an error if the symlink target doesn't exist.  So I went with:
```go
func (m *Mount) BaseDir() string {
        rawBaseDir := filepath.Join(m.Path, baseDirName)
        target, err := os.Readlink(rawBaseDir)
        if err != nil {
                return rawBaseDir // not a symlink
        }
        if filepath.IsAbs(target) {
                return target
        }
        return filepath.Join(m.Path, target)
}
```

One thing I don't quite like about either approach is that it causes `BaseDir()` (which is called several times on each mount) to do syscalls rather than just return a string. The single `readlink()` isn't too bad in comparison to the full path walk done by `EvalSymlinks()` though, so I suppose that for simplicity in the code we should just go with that for now.

filesystem/filesystem.go
ebiggers: Done

filesystem/path.go
ebiggers: It's still useful to have `isSymlink()` for the test, so I've kept it.

filesystem/filesystem_test.go
ebiggers: Like in the test from https://github.com/google/fscrypt/pull/152, it's just supposed to be finding a temporary path that doesn't exist yet.  I've changed it to using a path inside the TempDir() instead.

ebiggers:
@josephlr, any more comments on this pull request?

filesystem/filesystem.go
josephlr: This is much easier for me to understand. If it turns out `BaseDir()` making syscalls is a bottleneck for the `fscrypt` commands, we can always just cache the result (so that `readLink` is only ever called once per mount).

Issue #151, opened 2019-09-28
Auto unlock not working


Vermaak5:
I used these instructions (on arch) https://wiki.archlinux.org/index.php/Fscrypt to setup fscrypt for the home directory but the auto unlock feature does not work yet.
The encryption setup itself works fine but requires manual unlocking after login.

This is probably due to some pam issue. 

Here is  /etc/pam.d/system-login

```
#%PAM-1.0

auth       required   pam_tally2.so        onerr=succeed file=/var/log/tallylog
auth       required   pam_shells.so
auth       requisite  pam_nologin.so
auth       include    system-auth
auth       optional   pam_fscrypt.so

account    required   pam_tally2.so 
account    required   pam_access.so
account    required   pam_nologin.so
account    include    system-auth

password   include    system-auth

session    optional   pam_loginuid.so
session    optional   pam_keyinit.so       force revoke
session    include    system-auth
session    optional   pam_motd.so          motd=/etc/motd
session    optional   pam_mail.so          dir=/var/spool/mail standard quiet
-session   optional   pam_systemd.so
session    required   pam_env.so
session    optional   pam_fscrypt.so       lock_policies drop_caches

```

Here is /etc/pam.d/passwd

```
#%PAM-1.0
#password	required	pam_cracklib.so difok=2 minlen=8 dcredit=2 ocredit=2 retry=3
#password	required	pam_unix.so sha512 shadow use_authtok
password	required	pam_unix.so sha512 shadow nullok
password        optional        pam_fscrypt.so
```

There is also /etc/pam.d/fscrypt added with the latest release on arch see #130 

```
# Allow fscrypt to check your login passphrase when you create a login protector
auth		required	pam_unix.so
```

Any ideas where the issue might be?

ebiggers:
It works for me on a freshly set up Arch Linux system with exactly those PAM config files.  Just checking: when you ran `fscrypt encrypt`, did you make sure to choose the option to protect the directory with your login passphrase?  (If you chose "custom_passphrase", auto-unlocking won't work, even if the custom passphrase is the same as the login passphrase.)

Otherwise, you can try appending the `debug` parameter in `/etc/pam.d/system-login`, like:

    auth       optional   pam_fscrypt.so	debug

and 

    session    optional   pam_fscrypt.so    debug lock_policies drop_caches

Then after you log in check the log via `sudo journalctl -b | grep fscrypt` for anything unexpected.

Vermaak5:
The debug log showed the following message:

[...]

PR #152, opened 2019-09-29
actions/config: ensure config file is created with mode 0644

ebiggers:
If the user has set a restrictive umask, e.g. 0077, then
/etc/fscrypt.conf would be created without the world-readable bit set.
Fix it by overriding the umask when creating the file.

Resolves https://github.com/google/fscrypt/issues/151

### DIFF
--- a/actions/config.go
+++ b/actions/config.go
@@ -31,6 +31,7 @@ import (
 	"golang.org/x/sys/unix"
 
 	"github.com/google/fscrypt/crypto"
+	"github.com/google/fscrypt/filesystem"
 	"github.com/google/fscrypt/metadata"
 	"github.com/google/fscrypt/util"
 )
@@ -68,7 +69,8 @@ var (
 func CreateConfigFile(target time.Duration, useLegacy bool) error {
 	// Create the config file before computing the hashing costs, so we fail
 	// immediately if the program has insufficient permissions.
-	configFile, err := os.OpenFile(ConfigFileLocation, createFlags, configPermissions)
+	configFile, err := filesystem.OpenFileOverridingUmask(ConfigFileLocation,
+		createFlags, configPermissions)
 	switch {
 	case os.IsExist(err):
 		return ErrConfigFileExists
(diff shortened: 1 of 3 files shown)


actions/config.go
```diff
@@ -68,7 +68,9 @@ var (
 func CreateConfigFile(target time.Duration, useLegacy bool) error {
 	// Create the config file before computing the hashing costs, so we fail
 	// immediately if the program has insufficient permissions.
+	oldMask := unix.Umask(0)
 	configFile, err := os.OpenFile(ConfigFileLocation, createFlags, configPermissions)
+	unix.Umask(oldMask)
```
josephlr: I initially though this should be a `defer` in case `os.OpenFile` panics, but now I realize it's OK, as we shouldn't open any additional files on panic.

actions/config_test.go
```diff
@@ -0,0 +1,56 @@
...
+	oldMask := unix.Umask(0)
+	defer unix.Umask(oldMask)
+	unix.Umask(0077)
+
+	tempFile, err := ioutil.TempFile("", "fscrypt_conf")
+	if err != nil {
+		t.Fatal(err)
+	}
+	ConfigFileLocation = tempFile.Name()
+	os.Remove(ConfigFileLocation)
```
josephlr: Do we need this `os.Remove` and the `defer os.Remove` below?

It seems like it might be better to just use `ioutil.TempDir` and then cleanup everything in a single call.

actions/config_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+	err = CreateConfigFile(time.Millisecond, false)
```
josephlr: Combine this line with the line below, for style consistancy.

actions/config.go
ebiggers: I decided to change it to just do the OpenFile() in a helper function that overrides umask.

actions/config_test.go
ebiggers: Good idea.  I changed it to do that instead.

actions/config_test.go
ebiggers: Done

ebiggers:
@josephlr, any more comments on this pull request?

PR #159, opened 2019-10-25
Added capacity to slice creation, when capacity is known

vivekv96:
Simple optimisation to reduce memory allocations and copying(while appending).


vivekv96:
@josephlr, I found this by reading the code, but I am thinking of writing a small script over the weekend that can help with these kind of micro optimisations.
Something that probably does a grep on open source Go projects and finds slices that are being used immediately without a capacity, not sure how the script should work exactly but should be a fun side project.

PR #160, opened 2019-10-25
Format variable declarations to maintain consistency

AdityaHarindar:
Prefer variable declarations adhering to go style, `var a int` over `a := 0`

### DIFF
--- a/actions/callback.go
+++ b/actions/callback.go
@@ -85,7 +85,7 @@ func getWrappingKey(info ProtectorInfo, keyFn KeyFunc, retry bool) (*crypto.Key,
 // wrapping key until the correct key is returned by the callback or the
 // callback returns an error.
 func unwrapProtectorKey(info ProtectorInfo, keyFn KeyFunc) (*crypto.Key, error) {
-	retry := false
+	var retry bool
 	for {
 		wrappingKey, err := getWrappingKey(info, keyFn, retry)
 		if err != nil {
--- a/actions/policy.go
+++ b/actions/policy.go
@@ -47,7 +47,8 @@ var (
 // keyring. In order for this removal to have an effect, the filesystem should
 // also be unmounted.
 func PurgeAllPolicies(ctx *Context) error {
-	if err := ctx.checkContext(); err != nil {
+	var err error
+	if err = ctx.checkContext(); err != nil {
 		return err
 	}
 	policies, err := ctx.Mount.ListPolicies()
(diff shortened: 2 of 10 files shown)


josephlr:
What style guide is this from, I'm not sure what's wrong with `x := 0`. [Effective Go](https://golang.org/doc/effective_go.html) seems to prefer `:= 0`

AdityaHarindar:
As far as I can see, effective go prescribes `var x []<type>` over `x := []<type>` in the case of slices, as `var` would create a `nil` valued slice, while `:=` would create a non-nil slice of zero length. Also, general readability aside, I gave this PR to keep variable declarations consistent with the rest of the codebase. You can close this PR if you feel this change is not necessary.

josephlr:
> As far as I can see, effective go prescribes `var x []<type>` over `x := []<type>` in the case of slices, as `var` would create a `nil` valued slice, while `:=` would create a non-nil slice of zero length. Also, general readability aside, I gave this PR to keep variable declarations consistent with the rest of the codebase. You can close this PR if you feel this change is not necessary.

I think consistency for the codebase is a good thing. Ideally, we would use `name := value` when we are initializing a variable that's about to be used, and `var name Type` when we just need the variable to be _declared_, but when we are assigning to it later, say something like:
```go
var err error
if foo.Bar, err := Baz(); err != nil {
    // Error checking for Baz
}
// Use foo with Bar initialized
```

I tried to be consistent about this when writing, but I definitely missed a few. If you want to change this PR to fix them, that would be great! But it's not crucial, so if you just want to close this, it's also fine.

josephlr:
@AdityaHarindar most of he changes proposed here don't follow the rule of:
  - Use `:=` if you are declaring a variable and using it's value
  - Use `var name Type` if you are declaring a variable but are not using it's value (i.e. are assigning to it later)

There's also the issue that a lot of these changes just make for longer code for no gain. We should prefer code like:
```go
if err := Foo(); err != nil {
	// Handle error
}
```
instead of code like
```go
var err error
if err = Foo(); err != nil {
	// Handle error
}
```
The only time we want to have the `err` declared out of line is my example above: https://github.com/google/fscrypt/pull/160#issuecomment-546546648. This is because the following code is not valid Go:
```go
if foo.Bar, err := Baz(); err != nil {
    // Error checking for Baz
}
// Use foo with Bar initialized
```

I'm going to close this, but thanks for the PR.

Issue #164, opened 2019-11-14
Auto-generate recovery passphrase when protector is on a different filesystem


Redsandro:
@ebiggers [commented](https://github.com/google/fscrypt/issues/115#issuecomment-553059348):
> Login protectors are stored in the `/.fscrypt` directory on the __root filesystem.__ If that directory has been lost, then it's impossible to decrypt any directories that were protected by the login passphrase

Would it make sense to put this on the root of the __partition (device) that holds the encrypted directory__ in stead? If you have three drives with encrypted directories, and only the one that contains root file system breaks, it would mean data on _all_ drives is lost, not just the root drive that didn't even use encryption itself.

This actually happened to me on the [aforementioned](https://github.com/google/fscrypt/issues/115#issuecomment-553046403) computer. The SSD holding the root filesystem was from 2014 (5 years old). I do backup most of my files, but I do _not_ backup the root file system because I figure free software is already backed up on many crazy fast repositories (For example, I'm lucky to have gigabit internet and my ISP hosts a mirror of many repositories.) and software _configurations_ are stored in my home, which is backed up.

It's slightly frustrating to have two disks of data that I can't recover because some tiny metadata is not stored on the same device as the data that depends on it.

---

When part of my data - for example my `/home` directory - is on a separate partition, I already need to `fscrypt setup /home.` Now there is a `.fscrypt` directory on the root of both my `/` and my `/home/` partition.

I don't know how this works technically so I can only make assumptions here. But the `/.fscrypt` contains a `protector`, and the `/home/.fscrypt` contains a `protector.link`. If somehow the partition that contains the encrypted data links to the partition containing the protector, I am wondering if these could be reversed. Or let every partition that is set up for fscrypt keep their own copy of the protector. So that a perfectly healthy drive will never have to suffer data loss through the failure of a different drive.

ebiggers:
There are several disadvantages to doing this:

- If a user has N login protectors rather than 1, it would take N times longer for them to log in (assuming that the password hashing isn't made less secure to compensate).
- If the user changes their login passphrase and some filesystems on which they have login protectors are not currently mounted, then those login protectors wouldn't be updated.  So those directories would stop being auto-unlocked, as they would still be protected by the old passphrase.
[...]

PR #167, opened 2019-11-22
Automatically generate recovery passphrase when useful

ebiggers:
If a user re-installs their system (or otherwise loses the /.fscrypt
directory on the root filesystem) they also lose access to any login
passphrase-protected directories on other filesystems, unless additional
protectors were manually added.  This can be unexpected, as it may be
expected that the old login passphrase would still work.

We can't really fix this by storing a login protector on every
filesystem because:

- If a user were to have N login protectors, it would take them N times
  longer to log in, as every login protector would need to be unlocked.

- If a user were to change their login passphrase while any external
  volumes were unmounted, login protectors would get out of sync.

- It's preferable that an external volume isn't unlockable by itself
  using only a login passphrase, as login passphrases are often weak.

Instead, generate a recovery passphrase when creating a login
passphrase-protected directory on a non-root filesystem.

The recovery passphrase is added as a custom_passphrase protector, thus
giving the policy two protectors: one pam_passphrase and one
custom_passphrase.  Then this passphrase is written to a file in the new
encrypted directory.  Writing the passphrase to a file here is okay
since it's encrypted, but it's obviously useless by itself; it's up to
the user to store this passphrase somewhere else if they need it.

Use a recovery passphrase instead of a "recovery code" that encodes the
policy key directly because a passphrase is more user-friendly: it can
safely be made much shorter than a key, and it works just like any other
fscrypt protector.  Also, it's not as critical to allow recovery when
the .fscrypt directory on the *same* filesystem is deleted.

Update https://github.com/google/fscrypt/issues/164

### DIFF
--- a/actions/recovery.go
+++ b/actions/recovery.go
@@ -1,0 +1,100 @@
+/*
+ * recovery.go - support for generating recovery passphrases
+ *
+ * Copyright 2019 Google LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may not
+ * use this file except in compliance with the License. You may obtain a copy of
+ * the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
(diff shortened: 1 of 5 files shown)


ebiggers:
This has been out for review for 2 months, and also the recovery passphrase generation is easy to change later if needed.  So I'm going ahead and merging this.  @josephlr please do take a look later when you have time, though.

PR #168, opened 2019-11-27
travis: set GO111MODULE=on for 'go get'

ebiggers:
Ensure that the environmental variable GO111MODULE is set to "on" when
running 'go get'.  This fixes a CI failure with Go 1.11 and 1.12.


ebiggers:
Merged to fix CI build.

Issue #187, opened 2020-01-27
fscrypt encrypt: login protectors do not need a name


bonfus:
I'm trying to encrypt a bunch of directories in my home.
I did the setup with

    # fscrypt setup
    # fscrypt setup /
    # fscrypt setup /home

and status reports:

```
# fscrypt status
filesystems supporting encryption: 2
filesystems with fscrypt metadata: 2

MOUNTPOINT  DEVICE          FILESYSTEM  ENCRYPTION     FSCRYPT
/           /dev/dm-0       ext4        supported      Yes
/boot       /dev/nvme0n1p1  vfat        not supported  No
/home       /dev/dm-2       ext4        supported      Yes
```

Encryption with password works on both filesystems, but PAM instead fails:

```
$ fscrypt encrypt --verbose Documents
2020/01/27 10:22:14 Reading config from "/etc/fscrypt.conf"
2020/01/27 10:22:14 creating context for user "pie"
2020/01/27 10:22:14 Documents is on ext4 filesystem "/home" (/dev/dm-2)
2020/01/27 10:22:14 ensuring Documents is an empty and readable directory
2020/01/27 10:22:14 ensuring Documents supports encryption and filesystem is using fscrypt
2020/01/27 10:22:14 creating policy for "Documents"
2020/01/27 10:22:14 keyringID(_uid.1000) = 452341505, <nil>
2020/01/27 10:22:14 KeyctlLink(452341505, -1) = <nil>
2020/01/27 10:22:14 keyringID(session) = 416713166, <nil>
2020/01/27 10:22:14 KeyctlSearch(416713166, keyring, _uid.1000) = 452341505, <nil>
2020/01/27 10:22:14 listing descriptors in "/home/.fscrypt/protectors"
2020/01/27 10:22:14 found 1 descriptor(s)
2020/01/27 10:22:14 successfully read metadata from "/home/.fscrypt/protectors/904470698a0d233f"
2020/01/27 10:22:14 listing descriptors in "/.fscrypt/protectors"
2020/01/27 10:22:14 found 0 descriptor(s)
2020/01/27 10:22:14 adding additional ProtectorOptions
Should we create a new protector? [y/N] y
The following protector sources are available:
1 - Your login passphrase (pam_passphrase)
2 - A custom passphrase (custom_passphrase)
3 - A raw 256-bit key (raw_key)
Enter the source number for the new protector [2 - custom_passphrase]: 1
2020/01/27 10:22:26 using source: pam_passphrase
[...]

PR #189, opened 2020-01-27
actions/recovery: ensure recovery passphrase is really custom_passphrase

ebiggers:
If the login protector was just created by the same 'fscrypt encrypt'
command, then policy.Context.Config.Source will be pam_passphrase.  This
needs to be overridden to custom_passphrase when creating the protector
for the recovery passphrase.

This fixes the following error:

    fscrypt encrypt: login protectors do not need a name

Resolves https://github.com/google/fscrypt/issues/187

### DIFF
--- a/actions/recovery.go
+++ b/actions/recovery.go
@@ -26,7 +26,18 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/google/fscrypt/crypto"
+	"github.com/google/fscrypt/metadata"
 )
+
+// modifiedContextWithSource returns a copy of ctx with the protector source
+// replaced by source.
+func modifiedContextWithSource(ctx *Context, source metadata.SourceType) *Context {
+	modifiedConfig := *ctx.Config
+	modifiedConfig.Source = source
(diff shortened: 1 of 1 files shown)


ebiggers:
This bug is causing problems for users, so I'm going ahead and merging this.  @josephlr please take a look later if you have time.

actions/recovery.go
```diff
@@ -26,8 +26,19 @@ import (
...
 )
 
+// modifiedContextWithSource returns a copy of ctx with the protector source
+// replaced by source.
+func modifiedContextWithSource(ctx *Context, source metadata.SourceType) *Context {
+	modifiedConfig := *ctx.Config
+	modifiedConfig.Source = source
+	modifiedCtx := *ctx
+	modifiedCtx.Config = &modifiedConfig
+	return &modifiedCtx
```
josephlr: This isn't great, but is probably unavoidable without restructuring the underlying library to just not use `Context`s.

I'm wish Go supported Cloning a struct, as this code becomes hard to maintain and easy to break.

Issue #186, opened 2020-01-26
Protector no longer exists on linked filesystem


hulpor:
When setting up fscrypt (current git version) I ran into several issues.

First when I run 'fscrypt encrypt /home' (home is a mountpoint/separate partition) I get the message

> generating recovery passphrase because protectors are on a differnet filesystem

But the protectors are on the same filesystem at /home/.fscrypt

The encrypt operation still seemed to work, as there was no error message.

But when I run 'fscrypt unlock /home/targetfolder' I get the message that the folder is not encrypted.

Running fscrypt status /home will show a protector for the target folder but also the message:

> protector no longer exists on linked filesystem

After setting everything up several times I installed the last release v0.2.5 and everything works fine out of the box.

Looks like the current git version is broken for this setup. Let me now if I can do anything to help find whats causing this.

ebiggers:
Can you please provide the full commands you ran and their output, along with the contents of `/proc/self/mountinfo`?  Some of the things you wrote don't make sense; for example, `fscrypt encrypt /home` will actually always fail if `/home` is really the root directory of a filesystem.  Also, `fscrypt status /home` should only list policies and protectors; it shouldn't show any folder paths.  It would also be helpful to know which type of protector you're using.  If you're using a pam_passphrase protector, it's indeed expected that it's a linked protector and that a recovery passphrase is generated.

hulpor:
The correct comamnd that I used was indeed 
> fscrypt encrypt /home/username


I will recreate the setup as soon as possible to get the correct error and full messages.

Meanwhile this https://github.com/google/fscrypt/issues/187 was also part of the error messages

ebiggers:
Okay, I found how to reproduce this, so I don't think I need more information.

The main problem here is a duplicate of https://github.com/google/fscrypt/issues/187.  However there are two additional issues here:

* When `fscrypt encrypt` fails, the output is not clear in some cases.
* When `fscrypt encrypt` fails, newly-created linked protectors are not deleted.

ebiggers:
Can you check whether https://github.com/google/fscrypt/pull/189 fixes this?  I'll try to improve handling of `fscrypt encrypt` failures too, but https://github.com/google/fscrypt/pull/189 will make `fscrypt encrypt` not fail in the first place.

hulpor:
Since I reproduced it anyway, here is the result:
```

fscrypt setup
Create "/etc/fscrypt.conf"? [Y/n] 
[...]

PR #192, opened 2020-01-28
Clean up policies and protectors on error

ebiggers:
Ensure that new policies and protectors always get deleted if `fscrypt encrypt` fails partway through.

Update https://github.com/google/fscrypt/issues/186

### DIFF
--- a/cmd/fscrypt/commands.go
+++ b/cmd/fscrypt/commands.go
@@ -197,6 +197,7 @@ func encryptPath(path string) (err error) {
 		if policy, err = getPolicyFromFlag(policyFlag.Value, ctx.TargetUser); err != nil {
 			return
 		}
+		defer policy.Lock()
 
 		if !skipUnlockFlag.Value {
 			if err = validateKeyringPrereqs(ctx, policy); err != nil {
@@ -213,12 +214,12 @@ func encryptPath(path string) (err error) {
 		}
 
 		protector, created, protErr := selectOrCreateProtector(ctx)
-		// Successfully created protector should be reverted on failure.
 		if protErr != nil {
 			return protErr
 		}
 		defer func() {
 			protector.Lock()
+			// Successfully created protector should be reverted on failure.
 			if err != nil && created {
 				protector.Revert()
 			}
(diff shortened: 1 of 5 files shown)


cmd/fscrypt/commands.go
```diff
@@ -191,12 +191,14 @@ func encryptPath(path string) (err error) {
 
 	var policy *actions.Policy
 	var recoveryPassphrase *crypto.Key
+	var recoveryProtector *actions.Protector
```
josephlr: Does the protector need to be at this outer scope, or can it be in the inner scope where we create it.

filesystem/filesystem.go
```diff
@@ -393,24 +393,43 @@ func (m *Mount) AddProtector(data *metadata.ProtectorData) error {
...
+		existingLinkedMnt, err := getMountFromLink(string(existingLink))
+		if err != nil {
+			return false, err
+		}
+		if existingLinkedMnt != dest {
+			return false, errors.Wrapf(ErrFollowLink, "link %q points to %q, but expected %q",
+				linkPath, existingLinkedMnt.Path, dest.Path)
+		}
+		return false, nil
+	}
```
josephlr: I'm trying to figure out when this error condition (the link exists but is "wrong) could actually happen.

Would it make sense to "fix" the protector here?

cmd/fscrypt/commands.go
ebiggers: Moved to inner scope.

filesystem/filesystem.go
ebiggers: It could happen if someone reinstalls their root filesystem and restores the `.fscrypt` directory.  While I'd like to provide a way to automatically fix protector links, I think that if we were going to do that we should do it properly, maybe providing a `fscrypt fix-protectors` command that would handle fixing all linked protectors on all filesystems.  This particular code is only handling adding a protector to a policy, so it's not really the right place to try to fix the protector.  It won't run if someone is simply trying to use an existing encrypted directory.

ebiggers:
> It would be nice if we could just make all the changes in memory and then Commit() the new state in one go, but that's probably too complex.

The filesystem operations can still fail when committing the new state.  So I don't think that would help.

PR #193, opened 2020-01-28
cmd/fscrypt/commands: allow disabling recovery passphrase

ebiggers:
While it's important to generate a recovery passphrase in the linked protector case to avoid data loss if the system is reinstalled, some people really don't want it (even though it can be safely ignored as it almost certainly has far more entropy than the login passphrase).
    
As a compromise, prompt for y/n before generating it, with default y.  Also, to allow disabling the recovery passphrase during noninteractive use, add a --no-recovery command-line option.
    
Update https://github.com/google/fscrypt/issues/186


ebiggers:
Due to a merge conflict, this is based on https://github.com/google/fscrypt/pull/192.

PR #197, opened 2020-02-11
Release version v0.3.0

josephlr:


josephlr:
Whoops, we should do 0.2.6

Issue #206, opened 2020-03-23
Build isn't reproducible


ericonr:
Hey there!

Several Linux distros (at least Arch, Debian and SUSE) are striving for reproducible builds, as a way to verify build systems and guarantee the legitimacy of build servers and whatnot. The build time injection done in the Makefile makes building fscrypt an unreproducible process.

The Arch Linux [PKGBUILD](https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/fscrypt) already applies a [patch](https://git.archlinux.org/svntogit/community.git/tree/trunk/reprobuild.patch?h=packages/fscrypt) to the Makefile to make it reproducible, and on Debian it's [flagged as unreproducible](https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/fscrypt.html). Fixing this issue upstream is probably the best way of widely making fscrypt a reproducible build.

josephlr:
@ericonr, thanks for the report. For the record, I think it should be an explicit goal of fscrypt's development to have a reproducible build. Toward that end, I've opened #207.

PTAL, and let me know if that is not enough to resolve this issue.

ericonr:
Well, that was fast! Thanks for the quick response! This is quite a useful piece of software.

Stay safe!

PR #208, opened 2020-03-24
Makefile: Use trimpath if avalible

josephlr:
Passing -trimpath makes the build entirely reproducible. See #206 

Running `make` on two different computers (both Arch Linux, `go1.14.1`) gives a `sha256sum` of:
```
e6874f16445f7130cb1f7db63fb50850804f49d9b35af25e0a0f89b091f09ab7  ./bin/fscrypt
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,10 @@ override GO_LINK_FLAGS += $(VERSION_FLAG) -extldflags "$(LDFLAGS)"
 override GO_FLAGS += --ldflags '$(GO_LINK_FLAGS)'
 # Always use Go modules
 export GO111MODULE = on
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -73,6 +73,10 @@ override GO_LINK_FLAGS += $(VERSION_FLAG) -extldflags "$(LDFLAGS)"
 override GO_FLAGS += --ldflags '$(GO_LINK_FLAGS)'
 # Always use Go modules
 export GO111MODULE = on
+# Use -trimpath if avalible
```
ebiggers: Typo: "avalible" => "available"

Issue #211, opened 2020-03-28
[Question] [lxc] possible to use fscrypt inside a Linux Container (lxc)


graysky2:
I'd like to use `fscrypt` within a Linux Container ([LXC](https://github.com/lxc/lxc))  to encrypt a directory therein.  I'm finding that the user space util, `fscrypt` does not like the fact that it is containerized.  For example, I start the container and ssh into it:

```
# fscrypt setup
Defaulting to policy_version 2 because kernel supports it.
Customizing passphrase hashing difficulty for this system...
Created global config file at "/etc/fscrypt.conf".
[ERROR] fscrypt setup: /: not a mountpoint
```

Is there something I can modify with my config or via passing a switch upon setup or is this something that needs to be handled within my LXC config?

josephlr:
Filesystem encryption certainly should work inside of containers. What are the contents of `/proc/self/mountinfo` inside the container when you're trying to run `fscrypt setup`?

This might be a bug in fscrypt, but I'm not sure.

graysky2:
Within the container:
```
% cat /proc/self/mountinfo
699 650 259:3 /sandbox/rootfs / rw,relatime master:337 - ext4 /dev/nvme0n1p3 rw
700 699 0:63 / /dev rw,relatime - tmpfs none rw,size=492k,mode=755
701 699 0:62 / /proc rw,nosuid,nodev,noexec,relatime - proc proc rw
702 703 0:62 /sys/net /proc/sys/net rw,nosuid,nodev,noexec,relatime - proc proc rw
703 701 0:62 /sys /proc/sys ro,nosuid,nodev,noexec,relatime - proc proc rw
704 701 0:62 /sysrq-trigger /proc/sysrq-trigger ro,nosuid,nodev,noexec,relatime - proc proc rw
705 699 0:64 / /sys rw,nosuid,nodev,noexec,relatime - sysfs sysfs rw
706 705 0:64 / /sys ro,nosuid,nodev,noexec,relatime - sysfs sysfs rw
707 706 0:64 / /sys/devices/virtual/net rw,relatime - sysfs sysfs rw
708 707 0:64 /devices/virtual/net /sys/devices/virtual/net rw,nosuid,nodev,noexec,relatime - sysfs sysfs rw
709 706 0:61 / /sys/fs/fuse/connections rw,nosuid,nodev,noexec,relatime master:477 - fusectl fusectl rw
[...]

PR #213, opened 2020-04-01
Allow fscrypt to work in containers

ebiggers:
Update the /proc/self/mountinfo parsing code to allow selecting a Mount
with Subtree != "/", i.e. a Mount not of the full filesystem.  This is
needed to allow fscrypt to work in containers, where the root of the
filesystem may not be mounted.

See findMainMount() for details about the algorithm used.

Resolves https://github.com/google/fscrypt/issues/211

### DIFF
--- a/filesystem/filesystem.go
+++ b/filesystem/filesystem.go
@@ -70,8 +70,9 @@ var (
 //	DeviceNumber   - Device number of the filesystem.  This is set even if
 //			 Device isn't, since all filesystems have a device
 //			 number assigned by the kernel, even pseudo-filesystems.
-//	BindMnt        - True if this mount is not for the full filesystem but
-//			 rather is only for a subtree.
+//	Subtree        - The mounted subtree of the filesystem.  This is usually
+//			 "/", meaning that the entire filesystem is mounted, but
+//			 it can differ for bind mounts.
 //	ReadOnly       - True if this is a read-only mount
 //
 // In order to use a Mount to store fscrypt metadata, some directories must be
@@ -99,7 +100,7 @@ type Mount struct {
 	FilesystemType string
 	Device         string
 	DeviceNumber   DeviceNumber
-	BindMnt        bool
+	Subtree        string
 	ReadOnly       bool
 }
 
--- a/filesystem/mountpoint.go
+++ b/filesystem/mountpoint.go
@@ -39,6 +39,11 @@ import (
 
 var (
 	// This map holds data about the state of the system's filesystems.
+	//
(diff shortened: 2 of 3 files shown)


ebiggers:
@josephlr, are you planning to review this?

josephlr:
Hey, sorry for not looking at this sooner. My main concern with any change like this is that we would be _too_ permissive and we would put the `.fscrypt` directory at a bad location, potentially confusing users or leading to data loss.

What about the following algorithm (for when a device has multiple mounts)?
  1) If one of the mountpoints contains all the other mountpoints, choose that one
  2) If one of the subtrees contains all the other subtrees, choose that one

Note: For all of these (as before) we prefer R/W to R/O if things are otherwise identical.

If neither (1) nor (2) identity a unique mount (or they disagree), we say the situation is ambiguous. This also seems easier to explain/document, and is still a strict extension of the existing behavior, as only picking non-bind mounts (i.e. Subtree  = "/") is a subset of (2). 


With this simplification, all the tests would stay passing except for `TestLoadContainedSubtreesAndMountpoints`. But looking at mountinfo for that test:
```
222 15 259:3 /0   /tpm/a/b rw shared:1 - ext4 /dev/root rw
222 15 259:3 /1   /tpm/a   rw shared:1 - ext4 /dev/root rw
222 15 259:3 /2   /tpm/a/c rw shared:1 - ext4 /dev/root rw
222 15 259:3 /1/3 /tpm/d   rw shared:1 - ext4 /dev/root rw
222 15 259:3 /2/4 /tpm/e   rw shared:1 - ext4 /dev/root rw
```
It's not clear to me that any of those are the "main" mountpoint (despite the algorithm choosing `/1`).

Would we lose anything by implementing this simplification?

ebiggers:
The problem is that if someone had independent subtrees in one mountpoint tree (like the first three entries above, or like the example from https://github.com/google/fscrypt/issues/211 but rooted somewhere other than `/`), and then they bind-mounted some file or directory in that mountpoint tree to outside it (like entries 4-5 above), then with your proposal `fscrypt` would suddenly break as it would suddenly consider the mountpoints to be ambiguous.  I don't think that's desirable.  Instead, we should detect that the additional mounts are just bind mounts that are contained in what we had before.  (In the above example, entries 4 and 5 are contained in 2 and 3 respectively.)  Does that make sense?

josephlr:
@ebiggers that makes sense to me. We take the mounts, discard contained subtrees, then find the the mountpoint that contains all remaining mounts (if it exists).

Thanks again for dealing with the complexity in fixing this!!

Issue #215, opened 2020-04-01
Try to detect incomplete locking of v1-encrypted directory


ebiggers:
Unlike v2, `fscrypt lock` on a v1-encrypted directory doesn't warn about in-use files, as the kernel doesn't provide a way to easily detect it.  In https://github.com/google/fscrypt/issues/214 this caused some confusion.

It might be possible to implement a heuristic where we check whether the filenames in the directory seem to be no-key names or not.  If yes, the directory is fully locked, otherwise it's incompletely locked (or unlocked).

PR #217, opened 2020-04-19
Try to detect incomplete locking of v1-encrypted directory

ebiggers:
'fscrypt lock' on a v1-encrypted directory doesn't warn about in-use
files, as the kernel doesn't provide a way to easily detect it.

Instead, implement a heuristic where we check whether a subdirectory can
be created.  If yes, then the directory must not be fully locked.

Make both 'fscrypt lock' and 'fscrypt status' use this heuristic.

Resolves https://github.com/google/fscrypt/issues/215


ebiggers:
This pull request is now on top of the CLI tests pull request (https://github.com/google/fscrypt/pull/218).

PR #221, opened 2020-05-05
Fix build on Raspbian

fstanis:
Currently, trying to build on Raspbian fails with the following error:

```
$ go get github.com/google/fscrypt/cmd/fscrypt
# github.com/google/fscrypt/keyring
go/src/github.com/google/fscrypt/keyring/fs_keyring.go:231:6: constant 3225445912 overflows int
go/src/github.com/google/fscrypt/keyring/fs_keyring.go:235:7: constant 3225445913 overflows int
```

### DIFF
--- a/keyring/fs_keyring.go
+++ b/keyring/fs_keyring.go
@@ -228,7 +228,7 @@ func fsRemoveEncryptionKey(descriptor string, mount *filesystem.Mount,
 		return err
 	}
 
-	ioc := unix.FS_IOC_REMOVE_ENCRYPTION_KEY
+	ioc := uintptr(unix.FS_IOC_REMOVE_ENCRYPTION_KEY)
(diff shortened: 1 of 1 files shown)


keyring/fs_keyring.go
```diff
@@ -228,7 +228,7 @@ func fsRemoveEncryptionKey(descriptor string, mount *filesystem.Mount,
 		return err
 	}
 
-	ioc := unix.FS_IOC_REMOVE_ENCRYPTION_KEY
+	ioc := uintptr(unix.FS_IOC_REMOVE_ENCRYPTION_KEY)
```
ebiggers, line 231: This makes the cast to `uintptr` below redundant.  Can you remove that one?

ebiggers:
Updated and merged as c5764330ebc69f15b9bf13b94cfef45732864b5a.  Thanks!

PR #224, opened 2020-05-10
Declare instead of define variable in header file.

Etrnls:


ebiggers:
Thanks, this looks good.  I'm guessing you're using gcc 10 (which was just released), and this bug broke the build because gcc 10 defaults to `-fno-common`?  If so, can you mention that in the commit message?

Etrnls:
Yes this is to fix build with gcc 10.
Updated commit message.

ebiggers:
Merged, thanks!

PR #230, opened 2020-05-17
Release version v0.2.8

ebiggers:


ebiggers:
I pushed out the tag and added the release notes.

PR #234, opened 2020-06-03
travis.yml: build 32-bit binary

ebiggers:


ebiggers:
Combined with https://github.com/google/fscrypt/pull/235

Issue #233, opened 2020-06-02
fscrypt fails to build on 32bit x86: mismatched types int32 and int64


paride:
fscrypt v0.2.8 fails to build on i386 with the following error:
```
src/github.com/google/fscrypt/cmd/fscrypt/errors.go:121:19: invalid operation: statfs.Bsize != pagesize (mismatched types int32 and int64)
```
As i386 is a Debian release architecture this bug prevents the migration of the newer fscrypt versions to Debian testing.

ebiggers:
Thanks for reporting this; sorry for the trouble.  I should have tried to get a 32-bit build target in Travis CI the last time we ran into an issue like this.  Trying it now...

ebiggers:
Fixed by https://github.com/google/fscrypt/pull/235/commits/3b075f2fda880256b3d9de2e6197a224adc0a39f.  Also, https://github.com/google/fscrypt/commit/3c15a7a05bd7673d2ca4650dd1525df408d20728 added a 32-bit build to Travis CI, so this shouldn't happen again.

@josephlr, it might be necessary to release v0.2.9 already, to get this fix out.  Otherwise some people still can't upgrade from v0.2.5, unfortunately!

In the mean time, @paride, please feel free to carry the fix in the Debian package for v0.2.8.

ebiggers:
I've released v0.2.9 which includes this fix.

PR #256, opened 2020-10-11
travis.yml: upgrade to Ubuntu 20.04 and enable cli tests

ebiggers:
Now that Travis CI supports a version of Ubuntu that has a kernel that
supports v2 encryption policies, upgrade to it and enable the cli tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 sudo: false
-dist: bionic
+dist: focal
 arch:
   - amd64
   - ppc64le
@@ -65,6 +65,14 @@ jobs:
       script:
         - make coverage.out
         - goveralls -coverprofile=coverage.out -service=travis-ci
+
+    - name: Command-line interface tests
+      sudo: required
+      before_install:
+        - sudo apt-get -y install e2fsprogs expect keyutils
+      script:
+        - GO111MODULE=on go get github.com/google/fscrypt/cmd/fscrypt
+        - make cli-test
 
     - stage: deploy
       name: Release Binaries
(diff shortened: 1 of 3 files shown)


Maryse47:
did you considered migrating to [github actions](https://github.com/features/actions) instead?

josephlr:
@ebiggers can you update the expected CLI output to match what we're actually get. It looks like there is a disparity w.r.t the password being echoed.

We should proably also just ditch the 32-bit build. I doubt there's much value there, and we will probably have to continuously fight with the CI to prevent things from breaking.

> did you considered migrating to [github actions](https://github.com/features/actions) instead?

Yes, this is the long-term plan, once we get things working w/ 20.04 in our current setup.

josephlr (approved):

ebiggers:
> @ebiggers can you update the expected CLI output to match what we're actually get. It looks like there is a disparity w.r.t the password being echoed.

The problem is that the password is echoed when Travis CI runs the tests, but it isn't echoed when running the tests locally.  It presumably has something to do with whether stdin is a terminal or not.  Before we can merge this I need to find a way to make the output consistent.

> We should proably also just ditch the 32-bit build. I doubt there's much value there, and we will probably have to continuously fight with the CI to prevent things from breaking.

People still expect to build 32-bit binaries for ARM (https://github.com/google/fscrypt/pull/221), or for i386 (https://github.com/google/fscrypt/issues/233) which is still a supported Debian release architecture.

ebiggers:
> The problem is that the password is echoed when Travis CI runs the tests, but it isn't echoed when running the tests locally. It presumably has something to do with whether stdin is a terminal or not. Before we can merge this I need to find a way to make the output consistent.

Finally figured this out, I think.  getPassphraseKey() needs to put the terminal into raw mode before printing the prompt; otherwise there's a race condition.

Issue #260, opened 2020-10-28
Bash completion for the fscrypt command


audeoudh:
fscrypt lacks of a bash completion script.

I have written such a completion script, compatible with [scop/bash-completion](https://github.com/scop/bash-completion) toolkit.  According to their contributing guidelines, “before submitting a completion to us, first consider submitting it to the project that ships the commands your completion is for”.

Is the google/fscrypt's team interested in such a script?

josephlr:
Yes we would be interested, I'm wondering if it's possible to generate the autocompletions directly from the source code.

audeoudh:
Generate completions (i.e. run the program and get completions from it) or generate completion script, used during completion?

I do not know automated tools for generating completion scripts.  Some parts are really specialized, such as code that extracts the possible mountpoints or protector / policy IDs; some are general, and may be more easily extracted from it, such as the list of commands.

ebiggers:
Done by https://github.com/google/fscrypt/pull/261.  Thanks!

PR #261, opened 2020-10-30
bash-completion: add completion script

audeoudh:
Confer #260.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,6 +110,7 @@ lint: $(BIN)/golint $(BIN)/staticcheck $(BIN)/misspell
 	go list ./... | xargs -L1 golint -set_exit_status
 	staticcheck ./...
 	misspell -source=text $(FILES)
+	shellcheck -s bash cmd/fscrypt/fscrypt_bash_completion
 	( cd cli-tests && shellcheck -x *.sh)
 
 clean:
@@ -158,8 +159,8 @@ coverage.out: $(BIN)/gocovmerge $(COVERAGE_FILES)
 	@go test -coverpkg=./... -covermode=count -coverprofile=$@ -p 1 ./$* 2> /dev/null
 
 ###### Installation Commands (require sudo) #####
-.PHONY: install install-bin install-pam uninstall
-install: install-bin install-pam
+.PHONY: install install-bin install-pam uninstall install-completion
+install: install-bin install-pam install-completion
 
 PREFIX := /usr/local
 BINDIR := $(PREFIX)/bin
@@ -181,10 +182,16 @@ install-pam: $(PAM_MODULE)
 	install -d $(DESTDIR)$(PAM_CONFIG_DIR)
 	install $(PAM_CONFIG) $(DESTDIR)$(PAM_CONFIG_DIR)/$(NAME)
 
+COMPLETION_INSTALL_DIR := $(PREFIX)/share/bash-completion/completions
+
+install-completion: cmd/fscrypt/fscrypt_bash_completion
+	install -Dm644 $< $(DESTDIR)$(COMPLETION_INSTALL_DIR)/fscrypt
+
 uninstall:
 	rm -f $(DESTDIR)$(BINDIR)/$(NAME) \
 	      $(DESTDIR)$(PAM_INSTALL_PATH) \
-	      $(DESTDIR)$(PAM_CONFIG_DIR)/$(NAME)
+	      $(DESTDIR)$(PAM_CONFIG_DIR)/$(NAME) \
+	      $(DESTDIR)$(COMPLETION_INSTALL_DIR)/fscrypt
 
 #### Tool Building Commands ####
 TOOLS := $(addprefix $(BIN)/,protoc golint protoc-gen-go goimports staticcheck gocovmerge misspell)
--- a/cmd/fscrypt/fscrypt_bash_completion
+++ b/cmd/fscrypt/fscrypt_bash_completion
@@ -1,0 +1,286 @@
+# fscrypt_bash_completion
+#
+# Copyright 2017 Google Inc.
+# Author: Henry-Joseph Audéoud (h.audeoud@gmail.com)
+#
+# Licensed under the Apache License, Version 2.0 (the "License"); you may not
+# use this file except in compliance with the License. You may obtain a copy of
+# the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+# WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under
+# the License.
+
+
+# Prefer completion script style COMPREPLY=($(...)) assignment.
+# shellcheck disable=SC2207
+true  # To apply shellcheck directive to all file
+
+
+# Output list of possible mount points
+_fscrypt_mountpoints()
+{
+    # shellcheck disable=SC2016
+    fscrypt status 2>/dev/null | \
+        command awk 'substr($0, 1, 1) == "/" && $5 == "Yes" { print $1 }'
+}
+
+
+# Complete with all possible mountpoints
+_fscrypt_complete_mountpoint()
+{
+    COMPREPLY=($(compgen -W "$(_fscrypt_mountpoints)" -- "${cur}"))
+}
+
+
+# Output list of possible policy or protector IDs
+# $1: the mount point on which policies are looked for.
+# $2: the section (policy or protector) to retrieve
+_fscrypt_status_section()
+{
+    local section=${2^^}
+    # shellcheck disable=SC2016
+    fscrypt status "$1" 2>/dev/null | \
+        command awk '/^[[:xdigit:]]{16}/ && section == "'"$section"'" { print $1; next; }
+                     { section = $1 }'
+}
+
+
+# Complete with policies or protectors
+_fscrypt_complete_policy_or_protector()
+{
+    local status_section="$1"
+    if [[ $cur = *:* ]]; then
+        # Complete with IDs of the given mountpoint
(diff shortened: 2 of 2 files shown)


util/fscrypt_completion
```diff
@@ -0,0 +1,225 @@
+# bash completion for fscrypt                              -*- shell-script -*-
```
ebiggers, line 1: Can you add an Apache license header, like the other source code files in this project?

util/fscrypt_completion
```diff
@@ -0,0 +1,225 @@
...
+
+# Output list of possible mount points
+_fscrypt_mountpoints()
+{
+    fscrypt status 2>/dev/null | \
+        command sed \
+            -e '/^\//!d' \
+            -e '/Yes$/!d' \
```
ebiggers, line 9: This doesn't check which field the "Yes" is actually in.  It will match on a mountpoint `/mnt/Yes`, for example.

How about using:

            fscrypt status 2>/dev/null \
                 | awk 'substr($0, 1, 1) == "/" && $5 == "Yes" { print $1 }'

util/fscrypt_completion
```diff
@@ -0,0 +1,225 @@
...
+            -e 's/ .*//'
+}
+
+# Output list of possible policy or protector IDs
+# $1: the mount point on which policies are looked for.
+# $2: the section (policy or protector) to retrieve
+_fscrypt_status_section()
+{
+    fscrypt status "$1" 2>/dev/null | \
```
ebiggers, line 18: I'm struggling to understand these sed expressions.  I think the following would be easier to understand:

    local section=${2^^}
    fscrypt status "$1" 2>/dev/null \
            | awk '/^[[:xdigit:]]{16}/ && section == "'$section'" { print $1; next; }
                   { section = $1 }'

It also doesn't output extra blank lines.

util/fscrypt_completion
```diff
@@ -0,0 +1,225 @@
...
+}
+
+_fscrypt()
+{
+    local cur prev words cword split
+    local GLOBAL_OPTS="--verbose --quiet --help"
+    _init_completion -s -n : || return
+
+    case $prev in
+        --drop-caches)
```
ebiggers, line 32: `--drop-caches` is a no-op except for v1 encryption policies, which are deprecated.  So `--drop-caches` is effectively deprecated too.  I recommend omitting it from the bash completion support.

util/fscrypt_completion
```diff
@@ -0,0 +1,225 @@
...
+            -e "/^${2^^} /",'/^$/!d' \
+            -e "s/${2^^}//" \
+            -e "s/\s.*//"
+}
+
+_fscrypt()
+{
+    local cur prev words cword split
+    local GLOBAL_OPTS="--verbose --quiet --help"
+    _init_completion -s -n : || return
```
ebiggers, line 29: Can you add a brief comment about what `_init_completion()` does and what the `s` and `-n` options mean?  I can find it defined in `/usr/share/bash-completion/bash_completion`, but it was not at all obvious.

util/fscrypt_completion
```diff
@@ -0,0 +1,225 @@
...
+
+    case $prev in
+        --drop-caches)
+            COMPREPLY=($(compgen -W "true false" -- "${cur}"))
+            return ;;
+        --key)
+            _filedir
```
ebiggers, line 36: Is there any way to restrict the completion to nondirectory files only?

util/fscrypt_completion
```diff
@@ -0,0 +1,225 @@
...
+        --time)
+            # It's a time, hard to complete a number…
+            return ;;
+        --user)
+            COMPREPLY=($(compgen -u -- "${cur}"))
+            return ;;
+    esac
+
+    local -a positional
+    local cword cmd
```
ebiggers, line 70: The `cmd` variable is unused.

BTW, I found this by running `shellcheck`.  Can you please run `shellcheck` and fix any warnings it makes sense to fix?  At the moment all the other shell scripts in this project are `shellcheck`-clean, and `make lint` enforces this.  If there are some specific conventions for bash completion scripts that don't pass `shellcheck`, we can make an exception, but it would be preferable not to.

util/fscrypt_completion
```diff
@@ -0,0 +1,225 @@
+# bash completion for fscrypt                              -*- shell-script -*-
+
```
ebiggers, line 2: Also, this script doesn't belong in the `util/` directory.  There isn't a perfect place for it, but `cmd/fscrypt/` would be best I think.

ebiggers (commented):
Thanks, a few comments.  It's hard to review this without experience writing bash completion scripts, though.

util/fscrypt_completion
audeoudh: Sure.

util/fscrypt_completion
audeoudh: Well found!  I missed it.

util/fscrypt_completion
audeoudh: I'm struggling to understand these awk expressions, too. :-)  I may use yours if you prefer, it does not matter to me if I don't have to modify it.

util/fscrypt_completion
audeoudh: Yes, we shouldn't propose use of deprecate things.  I didn't know that `--drop-caches` was.

Dropped.

util/fscrypt_completion
audeoudh: `/usr/share/bash-completion/bash_completion` comes with the [scop/bash-completion](https://github.com/scop/bash-completion) toolbox.  That is a common way, in completion script context, to initialize completion variables thus.  Yes, I'll add a comment.

util/fscrypt_completion
audeoudh: You don't want that: directories may contains keys, and you want to complete these directories to be able to reach that keys.  Isn't it?

util/fscrypt_completion
audeoudh: Well catched.  I checked the file with the scop/bash-completion test suite, but I didn't use shellcheck.

I fixed most of them, and add a comment for all explicitly disabled warnings.  The one I cannot resolve is SC2016 on your awk expressions: `$` is wrongly recognized by shellcheck as a bash variable.  Do you have an idea on the best way to resolve it?

Should I add something in the Makefile for that file to be checked via `make lint`?

util/fscrypt_completion
audeoudh: Moved.

Should that script be installed as `$PREFIX/share/bash-completion/completions/fscrypt` while running `make install`?

util/fscrypt_completion
ebiggers: Yes, if possible shellcheck should be run by `make lint`.

util/fscrypt_completion
ebiggers: Yes, I think `make install` should install this.

ebiggers:
I think I'm finally understanding most of this, but it's difficult to read because of the code duplication with the bash-completion boilerplate.  Could you add some helper functions like:

```
_fscrypt_complete_mountpoint()                                                   
{                                                                                
    COMPREPLY=($(compgen -W "$(_fscrypt_mountpoints)" -- "$cur"))                
}

_fscrypt_complete_word()                                                         
{                                                                                
    local wordlist=$1                                                            
                                                                                 
    COMPREPLY=($(compgen -W "$wordlist" -- "$cur"))                              
}                                                                                
                                                                                 
_fscrypt_complete_option()                                                       
{                                                                                
    local additional_opts=${1-}                                                  
    local GLOBAL_OPTS="--verbose --quiet --help"                                 
                                                                                 
    _fscrypt_complete_word "$additional_opts $GLOBAL_OPTS"                       
}                                                                                
```

... and use them everywhere where they can.  Also, please use `if` instead of `case` when there are only two branches.

That should make things a lot easier to understand.

E.g. the following:

            case $cur in
                -*)
                    COMPREPLY=($(compgen -W "--time= --force ${GLOBAL_OPTS}" \
                               -- "${cur}"))
                    ;;
                *)
                    COMPREPLY=($(compgen -W "$(_fscrypt_mountpoints)" -- "${cur}"))
                    ;;
            esac

... would become:

            if [[ $cur == -* ]]; then
                _fscrypt_complete_option "--time= --force"
            else
                _fscrypt_complete_mountpoint
            fi

cmd/fscrypt/fscrypt_completion
```diff
@@ -0,0 +1,252 @@
+# bash completion for fscrypt                              -*- shell-script -*-
+#
+# fscrypt_completion - Custom errors and error functions used by fscrypt
```
ebiggers, line 3: This line seems to have been copy and pasted from another file. It should be removed.

ebiggers:
Also, please squash your commits into one.  Thanks!

audeoudh:
I mainly:
- changed “too simple” `case` structures to `if`/`else` blocks,
- added functions for conceptualize bash-completion typical code,
- added more comments in the script,
- added the script in `make lint` and `make install`.
- squashed commits

Let me know if there is still something to do.

util/fscrypt_completion
ebiggers: Looks like `--drop-caches` didn't get fully removed.

util/fscrypt_completion
ebiggers: Can you make `make uninstall` delete it, as well?

ebiggers:
It looks much better now, thanks.  Can you address the two comments I just left (remove `--drop-caches`, and update `make uninstall`)?  Afterwards I think this will be ready to merge, unless @josephlr has comments.

util/fscrypt_completion
audeoudh: Oops, I missed some of them…

util/fscrypt_completion
audeoudh: Done.

ebiggers:
Thanks, merged now.

PR #267, opened 2020-11-21
Switch from Travis CI to GitHub Actions

ebiggers:
travis-ci.org is being shut down, so switch to GitHub Actions.
    
It should be mostly equivalent, but I did drop functionality in a couple cases:
    
* Publishing release binaries.  I don't think providing Linux binaries is useful, since people build their own anyway.  So I left this out.
    
* Build and testing on ppc64le.  GitHub Actions only natively supports x86.  I tried uraimo/run-on-arch-action, which uses Docker and QEMU user-mode emulation, but the fscrypt tests can't be run because QEMU user-mode emulation doesn't support all the needed system calls.


ebiggers:
Note, GitHub Actions isn't actually running on this pull request yet.  I need to figure out why.  It was working for me on a test pull request on my fork (https://github.com/ebiggers/fscrypt/pull/1/checks).

josephlr:
> @josephlr, any idea why there is now also a Cirrus CI check, which is failing?

Ya, I was originally going to switch things to CirrusCI, but never got around to it. I've denied the hook, so let's see if this disables the check.

josephlr (approved):

PR #271, opened 2020-12-10
read raw key from stdin if --key isn't specified

myomv100:
Hello everyone,

This pull request contains some additions to fscrypt so you can now pass base64 encoded raw key and use it without the need to create the raw key file on the file system and worry about its security.

Best

### DIFF
--- a/cmd/fscrypt/keys.go
+++ b/cmd/fscrypt/keys.go
@@ -22,6 +22,8 @@
 package main
 
 import (
+	"bufio"
+	"bytes"
 	"fmt"
 	"io"
 	"log"
@@ -178,6 +180,19 @@ func makeKeyFunc(supportRetry, shouldConfirm bool, prefix string) actions.KeyFun
 			// Only use prefixes with passphrase protectors.
 			if prefix != "" {
 				return nil, ErrNotPassphrase
+			}
+			if keyFileFlag.Value == "" { // Reading from STDIN
+				in := bufio.NewReader(os.Stdin)
+				buf := make([]byte, 0, 32)
+				for i := 1; i <= 32; i++ {
+					c, err := in.ReadByte()
+					buf = append(buf, c)
+					if err == io.EOF {
+						break
+					}
+				}
+				key := bytes.NewReader(buf)
+				return crypto.NewFixedLengthKeyFromReader(key, metadata.InternalKeyLen)
 			}
 			prompt := fmt.Sprintf("Enter key file for protector %q: ", info.Name())
 			// Raw keys use a file containing the key data.
--- a/ThreatDragonModels/Test/Test.json
+++ b/ThreatDragonModels/Test/Test.json
@@ -1,0 +1,17 @@
+{
+  "summary": {
+    "owner": "Test",
+    "title": "Test"
+  },
+  "detail": {
+    "contributors": [],
+    "diagrams": [
+      {
+        "title": "Test",
+        "thumbnail": "./public/content/images/thumbnail.jpg",
+        "id": 0
+      }
+    ],
+    "reviewer": ""
+  }
+}


myomv100:
@googlebot I signed it!

cmd/fscrypt/commands.go
```diff
@@ -107,7 +107,7 @@ var Encrypt = cli.Command{
 		immediately be used.`, directoryArg, shortDisplay(policyFlag),
 		shortDisplay(protectorFlag), mountpointArg),
 	Flags: []cli.Flag{policyFlag, unlockWithFlag, protectorFlag, sourceFlag,
-		userFlag, nameFlag, keyFileFlag, skipUnlockFlag, noRecoveryFlag},
+		userFlag, nameFlag, keyFileFlag, keyEncodedStrFlag, skipUnlockFlag, noRecoveryFlag},
```
ebiggers, line 110: Command-line arguments shouldn't contain raw keys, as command lines are typically visible to other users on Linux.

The key will need to be passed on standard input instead (and then it can be binary rather than hex).  Take a look at what https://github.com/google/fscryptctl does.

actions/callback.go
```diff
@@ -65,7 +65,8 @@ type KeyFunc func(info ProtectorInfo, retry bool) (*crypto.Key, error)
 // passphrase sources or just relays the callback for raw sources.
 func getWrappingKey(info ProtectorInfo, keyFn KeyFunc, retry bool) (*crypto.Key, error) {
 	// For raw key sources, we can just use the key directly.
-	if info.Source() == metadata.SourceType_raw_key {
+	if info.Source() == metadata.SourceType_raw_key || 
+	   info.Source() == metadata.SourceType_raw_keyEncodedStr {
```
ebiggers, line 69: This shouldn't really be a separate type of protector, but rather just another way of specifying the raw key for a raw key protector.

ebiggers (changes requested):

cmd/fscrypt/commands.go
myomv100: It makes sense.  Since in my case it won't be used by normal users but by IPC handled by root I added the CLI argument to make it easier.. I removed the argument from the pull request. So now it has to be entered on the standard input.

actions/callback.go
myomv100: Exactly It's one thing , a raw key from different sources.. so we could say the sources of the raw key source. 

Last changes makes it easy to add more sources for raw key. For may case,, I need to do some sort of remote encryption/unlocking mechanism... maybe over ssh or any key management server api.. still don't know best way to do so. Pam might be the best place for the unlocking part but not the encryption itself...

ebiggers:
Sorry for the slow response.  A couple comments:

Storing a "raw key type" on disk feels like the wrong approach.  This is really just another way of passing a raw key to the program, so there shouldn't be any need to store anything on-disk.  Why not just make `fscrypt` read the key from standard input if it needs a raw key, is being run in noninteractive mode, and the `--key` argument isn't specified?

Also, please ensure that all tests are passing.  See CONTRIBUTING.md for how to run them yourself.  E.g., `make format` needs to be run.

myomv100:
> Sorry for the slow response. A couple comments:
> 
> Storing a "raw key type" on disk feels like the wrong approach. This is really just another way of passing a raw key to the program, so there shouldn't be any need to store anything on-disk. Why not just make `fscrypt` read the key from standard input if it needs a raw key, is being run in noninteractive mode, and the `--key` argument isn't specified?
> 
> Also, please ensure that all tests are passing. See CONTRIBUTING.md for how to run them yourself. E.g., `make format` needs to be run.

You mean we should remove the currently supported option of passing a raw key file that's stored on disk ?

Also writing to standard input a raw data will mess up bash environment. The terminals are not optimized to process binary data.   
 So `--key` shall gets the key from stdin. And to make sure we don't mess up the terminal, we need to remove it from the menu of raw key prompt in the interactive mode.

Is this what you mean?

ebiggers:
> You mean we should remove the currently supported option of passing a raw key file that's stored on disk ?

No, that's not necessary.  In noninteractive mode, if --key is specified then read the key from the specified file, otherwise read the key from stdin.

> Also writing to standard input a raw data will mess up bash environment. The terminals are not optimized to process binary data.

This is only true in interactive mode.

It would be possible to make the interactive mode ask whether the user wants to provide the key as a key file or as a base64-encoded string.  However, would this actually be useful?  I think the main benefit of supporting providing the key on stdin is for noninteractive mode.

myomv100:
> > You mean we should remove the currently supported option of passing a raw key file that's stored on disk ?
> 
> No, that's not necessary. In noninteractive mode, if --key is specified then read the key from the specified file, otherwise read the key from stdin.
> 
> > Also writing to standard input a raw data will mess up bash environment. The terminals are not optimized to process binary data.
> 
> This is only true in interactive mode.
> 
> It would be possible to make the interactive mode ask whether the user wants to provide the key as a key file or as a base64-encoded string. However, would this actually be useful? I think the main benefit of supporting providing the key on stdin is for noninteractive mode.

Last commit includes reading raw key file stdin  but you need to pass `--key=- ` so we know you want to read it from stdin . The dash character to read from stdin argument is barrowed from [cryptsetup ](https://gitlab.com/cryptsetup/cryptsetup)  for exactly the same purpose which is reading key files.

ebiggers:
> Last commit includes reading raw key from stdin but you need to pass --key=- so we know you want to read it from stdin .

Any reason not to just read the key from stdin if `--key` isn't specified, as I suggested?

Also, please drop any obsolete changes from the pull request (and squash together commits as needed).  There are still changes to add RawKeyType, but that shouldn't be needed as I explained earlier, right?

myomv100:
> > Last commit includes reading raw key from stdin but you need to pass --key=- so we know you want to read it from stdin .
> 
> Any reason not to just read the key from stdin if `--key` isn't specified, as I suggested?
> 
> Also, please drop any obsolete changes from the pull request (and squash together commits as needed). There are still changes to add RawKeyType, but that shouldn't be needed as I explained earlier, right?
 
Done !

cmd/fscrypt/keys.go
```diff
@@ -179,6 +181,19 @@ func makeKeyFunc(supportRetry, shouldConfirm bool, prefix string) actions.KeyFun
...
+				buf := make([]byte, 0, 32)
+				for i := 1; i <= 32; i++ {
+					c, err := in.ReadByte()
+					buf = append(buf, c)
+					if err == io.EOF {
+						break
+					}
+				}
+				key := bytes.NewReader(buf)
+				return crypto.NewFixedLengthKeyFromReader(key, metadata.InternalKeyLen)
```
ebiggers, line 195: Why not just:

    return crypto.NewFixedLengthKeyFromReader(bufio.NewReader(os.Stdin), metadata.InternalKeyLen)

ebiggers (changes requested):
Can you please update `cli-tests/t_encrypt_raw_key.sh` to test this new feature?

Also, this breaks interactive use of raw_key:

    $ fscrypt encrypt dir/
    Should we create a new protector? [y/N] y
    The following protector sources are available:
    1 - Your login passphrase (pam_passphrase)
    2 - A custom passphrase (custom_passphrase)
    3 - A raw 256-bit key (raw_key)
    Enter the source number for the new protector [2 - custom_passphrase]: 3
    Enter a name for the new protector: raw_key

The program just hangs after that because it is reading from stdin.

As I suggested earlier, reading the key from stdin should only happen in noninteractive mode.

dimitry-ishenko:
This seems so close to being done. I can try to push it over the hump, as I need this functionality.

@ebiggers is there a preferred way of detecting non-interactive mode in Go? I've found the [`go-isatty`](https://pkg.go.dev/github.com/mattn/go-isatty) package. Will that work?

ebiggers:
> @ebiggers is there a preferred way of detecting non-interactive mode in Go? I've found the go-isatty package. Will that work?

I think you can just use `term.IsTerminal()`, like `cmd/fscrypt/keys.go` uses.

dimitry-ishenko:
@ebiggers 👍🏻 👍🏻 👍🏻 Please see PR #326

Issue #272, opened 2020-12-23
Destroying an in-use protector should be forbidden?


johnsutton:
With v0.2.9 you can successfully "metadata destroy --protector X" even though the protector is currently in use with a policy, and it then becomes impossible to remove-protector-from-policy because "protector metadata for X not found on filesystem Y".

Arguably, either the destroy should be forbidden OR all uses of the protector should be removed at the same point.

And a small bugette: if you attempt to unlock a directory which is using a policy which is "protected" by such a non-existent protector (in addition to one or more real, existing protectors) then you get a printf output with uninstantiated %placeholders:

The available protectors are: 
1 - login protector for someuser
2 - custom protector "newprot"
NOTE: %d of the %d protectors failed to load. You may need to mount a linked filesystem. Run
with --verbose for more information.Enter the number of protector to use:

Other than this... great work!  Thanks :)

mahiarmody:
I completely agree. I'd like to offer additional possibilities too. In addition to to the above-mentioned options suggested:
1. Forbid the deletion of a protector in use, stating that it must first be removed from all policies using it.
OR
2. The protector should be automatically removed from all policies using it and this should be stated as a warning.

Also alternatively,
3. Provide a new subcommand that cleans off all such junk / dangling protector entries from all policies.
OR
4. Modify the existing `fscrypt metadata remove-protector-from-policy` command to ignore the error:
`scrypt metadata remove-protector-from-policy: protector metadata for <protector ID> not found on filesystem <mountpoint>` 
and proceed with removal of the deleted protector from the policy. After all a deleted protector can't be used any way, so why keep it?

josephlr:
@johnsutton @mahiarmody thanks for bringing this up. I agree that this workflow needs to be improved. I was able to get into a similar situation by manually deleting protector files in the `.fscrypt` metadata directory.

At the very least, we should allow `fscrypt metadata remove-protector-from-policy` to work even if the protector doesn't exist. The current functionality puts users in an unfixable state.

ebiggers:
https://github.com/google/fscrypt/pull/338 will make it possible to use `fscrypt remove-protector-from-policy` even if the protector no longer exists. 

I'm not sure that the command to destroy a protector needs to be changed too.  Regarding forbidding destroying an in-use protector: people might want to do this intentionally.  Regarding automatically finding and updating all policies: that would require a global scan, which would be preferable to avoid.  For now I'd just like to fix `fscrypt metadata remove-protector-from-policy`.

PR #275, opened 2021-01-15
cmd/fscrypt: fix missing protector error format

hobbitalastair:
This is the bugette mentioned in #272 - the format specifiers were never being filled in.
Also add a newline so the prompt to select a protector shows on a separate line.

### DIFF
--- a/cmd/fscrypt/prompt.go
+++ b/cmd/fscrypt/prompt.go
@@ -282,7 +282,8 @@ func promptForProtector(options []*actions.ProtectorOption) (int, error) {
 	}
 
 	if numLoadErrors > 0 {
-		fmt.Print(wrapText("NOTE: %d of the %d protectors failed to load. "+loadHelpText, 0))
+		loadWarning := fmt.Sprintf("NOTE: %d of the %d protectors failed to load. ", numLoadErrors, numOptions)
+		fmt.Print(wrapText(loadWarning+loadHelpText, 0) + "\n")
 	}
 
 	for {


hobbitalastair:
@googlebot I signed it!

cmd/fscrypt/prompt.go
```diff
@@ -282,7 +282,7 @@ func promptForProtector(options []*actions.ProtectorOption) (int, error) {
 	}
 
 	if numLoadErrors > 0 {
-		fmt.Print(wrapText("NOTE: %d of the %d protectors failed to load. "+loadHelpText, 0))
+		fmt.Print(wrapText(fmt.Sprintf("NOTE: %d of the %d protectors failed to load. ", numLoadErrors, numOptions)+loadHelpText, 0) + "\n")
```
ebiggers, line 285: This long line is difficult to read.  Can you break it up, like the following:

                warning := fmt.Sprintf("NOTE: %d of the %d protectors failed to load. ", numLoadErrors, numOptions)
                fmt.Print(wrapText(warning+loadHelpText, 0) + "\n")

ebiggers (commented):
Could you add `Update #272` to the commit message so that it directly links to the bug report?

hobbitalastair:
@ebiggers thanks for the review, good point about the long line

I've called the formatted text `loadWarning` to be more descriptive, happy to change if you would rather the shorter version.

ebiggers:
Merged, thanks!

Issue #270, opened 2020-12-05
pulseaudio fails to automatically start on login when the home directory is encrypted


mahiarmody:
My environment is the Ubuntu 20.04. I installed fscrypt after compiling from source as described on this GitHub site. After installing fscrypt I noticed that pulseaudio failed to automatically start after login. I had to explicitly start pulseaudio manually each time using:
`pulseaudio --start`

Here are my system details:

**Kernel version:**
5.4.0-56-generic

**Ubuntu details:**
NAME="Ubuntu"
VERSION="20.04.1 LTS (Focal Fossa)"
ID=ubuntu
ID_LIKE=debian
PRETTY_NAME="Ubuntu 20.04.1 LTS"
VERSION_ID="20.04"
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
VERSION_CODENAME=focal
UBUNTU_CODENAME=focal

**fscrypt version:**
fscrypt version v0.2.9-11-gcc933cd

**To fix this issue permanently, I had to make the following change:**

In file `/etc/pam.d/common-session` move the line containing `pam_systemd.so` below the line
containing `pam_fscrypt.so`.

So, in file: `/etc/pam.d/common-session` change the "Additional block" section like so:

FROM:
> \# and here are more per-package modules (the "Additional" block)
> session	required	pam_unix.so 
> session optional	pam_systemd.so
> `session	optional	/usr/lib/security/pam_fscrypt.so drop_caches lock_policies`
> \# end of pam-auth-update config

TO:
> \# and here are more per-package modules (the "Additional" block)
> session	required	pam_unix.so 
> `session	optional	/usr/lib/security/pam_fscrypt.so drop_caches lock_policies`
> session optional	pam_systemd.so
> \# end of pam-auth-update config

**I found this fix by referring to the following link:**
[https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1889416](https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1889416)

It would be great if this could be handled automatically at install time.

Thank you.

petergardfjall:
I can confirm the issue. I'm also running `Ubuntu 20.04.1 LTS`, `5.4.0-56-generic` (though I'm on Xubuntu).
[...]

PR #278, opened 2021-03-02
pam_fscrypt/config: prioritise over other session modules

ramcq:
Services launched by systemd user sessions on Debian / Ubuntu systems are often not able to access the home directory, because there is no guarantee / requirement that pam_fscrypt is sequenced before pam_systemd. Although this pam-config mechanism is Debian-specific, the config file is provided here upstream and unmodified in Debian. Raising the priority here so that it's always ordered ahead of pam_systemd will solve issues such as https://github.com/google/fscrypt/issues/270, https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=964951 and https://bugs.launchpad.net/ubuntu/+source/systemd/+bug/1889416.

### DIFF
--- a/pam_fscrypt/config
+++ b/pam_fscrypt/config
@@ -1,6 +1,6 @@
 Name: fscrypt PAM passphrase support
 Default: yes
-Priority: 0
+Priority: 100
 Auth-Type: Additional
 Auth-Final:
 	optional	PAM_INSTALL_PATH


ramcq:
@googlebot I signed it!

ramcq:
@googlebot I signed it!

ramcq:
@googlebot I signed it!

ramcq:
Sorry about spam. This it the most labour-intensive one-line PR I've ever sent. I've signed the corporate CLA... got the right e-mail address on my GitHub account... not sure what else is needed...?!

ebiggers:
> I've signed the corporate CLA... got the right e-mail address on my GitHub account... not sure what else is needed...?!

Did you follow https://cla.developers.google.com/about#add-contributors to add your email address as an authorized contributor for your company?  I think that might be the missing step.

ebiggers:
The actual fix looks good to me, although it would be helpful if there was some more reasoning behind the exact priority value that is being chosen (100).  It needs to be higher priority than pam_systemd (0) and lower priority than pam_unix (256), but are there any additional constraints?

Also, can you please wrap your commit message at 72 columns, as per the usual convention?

Thanks!

ramcq:
> > I've signed the corporate CLA... got the right e-mail address on my GitHub account... not sure what else is needed...?!
> 
> Did you follow https://cla.developers.google.com/about#add-contributors to add your email address as an authorized contributor for your company? I think that might be the missing step.

I may just have been too impatient - the corporate CLA has only come back to me from DocuSign today. Also (although both addresses are on my GitHub account) from an abundance of caution I've just made sure the commit author matched the current primary e-mail address on my Google account.

ramcq:
> The actual fix looks good to me, although it would be helpful if there was some more reasoning behind the exact priority value that is being chosen (100). It needs to be higher priority than pam_systemd (0) and lower priority than pam_unix (256), but are there any additional constraints?

I made a small survey of the pam-configs files present in Debian bullseye that include Session plugins: [session-priorities.txt](https://github.com/google/fscrypt/files/6075685/session-priorities.txt). Without auditing every PAM plugin to see what precisely is done, it seems that things which interact with / fetch network credentials and groups are first with very high priorities, followed by eg pam_mount at 128 which could be implementing a necessary site-specific policy to mount the home directory, and then some things at 64 such as pam_ssh (launches ssh-agent) which seem unlikely to function properly without the home directory already being available. So, 100 was a slightly arbitrary choice, but it doesn't seem an unreasonable one. Happy to amend if you feel differently, but I've added a brief note to the commit message along these lines.

> Also, can you please wrap your commit message at 72 columns, as per the usual convention?

Done!

> Thanks!

Welcome! Thanks for fscrypt. :)

ebiggers:
Merged, thanks for the fix!

I'll probably need to fix the directions for Arch Linux too (https://wiki.archlinux.org/index.php/Fscrypt#PAM_module).  They currently say to "append" pam_fscrypt, which would imply putting it after pam_systemd.

There is also the [Enabling the PAM module on other Linux distros](https://github.com/google/fscrypt#enabling-the-pam-module-on-other-linux-distros) section in the README.  It already says to put pam_fscrypt "after pam_unix", so it sort of recommends the right thing already, but it could be clarified.

PR #280, opened 2021-03-03
README: add note about ordering pam_fscrypt before pam_systemd

ramcq:
As discussed in https://github.com/google/fscrypt/pull/278#issuecomment-789946295

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,6 +425,15 @@ locked data is inaccessible; this only needed for v1 encryption policies.  All
 the types also support the `debug` option which prints additional debug
 information to the syslog.
 
+Note that in order for encrypted home directories to work correctly,
+`pam_fscrypt.so` must be placed _before_ any other PAM Session modules which
+rely on access to the home directory, or start processes which access your
+home directory during the duration of your session. Particularly, systems
+which use `systemd-logind` for managing jobs and cgroups in user sessions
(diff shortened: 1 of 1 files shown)


ebiggers:
Ah, I didn't see you were working on this.  I just opened a pull request too (https://github.com/google/fscrypt/pull/281/), which also does some other things such as improve the default behavior of `pam_fscrypt.so` so that the lock_policies and drop_caches options aren't needed.

I went with a briefer note about ordering, in-line with the sentence about where to add the line (so that it's harder to miss):

> Add the line ...  after `pam_unix.so` in `/etc/pam.d/common-session` or similar, but before `pam_systemd.so` or any other module that requires the user's home directory to be available.

Do you think that's sufficient, or could it use more explanation?

ramcq:
Seems alright, I might've gone a little more explicit on accesses the home directory or starts processes, as it's a little hidden / non-obvious that pam_systemd starts a session behind your back these days. :)

ebiggers:
Updated to:

> Add the line ... after `pam_unix.so` in `/etc/pam.d/common-session` or similar, but before `pam_systemd.so` or any other module that accesses the user's home directory or starts processes which access the user's home directory during their session.

ebiggers:
https://github.com/google/fscrypt/pull/281 was merged, which took care of this.

PR #283, opened 2021-03-31
Stop generating and uploading coverage in CI

josephlr:
This is currently broken, and we don't really use the findings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -64,12 +64,8 @@ jobs:
       run: |
         make test-setup
         keyctl link @u @s
-        make coverage.out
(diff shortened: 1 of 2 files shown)


ebiggers:
Should the badge at the top of the README be removed too?

josephlr:
> Should the badge at the top of the README be removed too?

Yes, good catch. Done!

josephlr:
I also added https://github.com/google/fscrypt/pull/283/commits/c3bb6c7d9304894a35a2229f4ceda76f745fc264 to avoid duplicate CI checks.

ebiggers (approved):

PR #284, opened 2021-04-17
Makefile: Optionally avoid installation of Ubuntu-specific PAM config

Flowdalic:
This allows non Ubuntu distributions to opt out from the installation of Ubuntu-specific PAM files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,7 @@ PAM_NAME := pam_$(NAME)
 # BINDIR: Where to install the fscrypt binary.        Default: $(PREFIX)/bin
 # PAM_MODULE_DIR: Where to install pam_fscrypt.so.    Default: $(PREFIX)/lib/security
 # PAM_CONFIG_DIR: Where to install Ubuntu PAM config. Default: $(PREFIX)/share/pam-configs
+#   If the empty string, then the Ubuntu PAM config will not be installed.
 #
 # MOUNT: The filesystem where our tests are run.    Default: /mnt/fscrypt_mount
 #   Ex: make test-setup MOUNT=/foo/bar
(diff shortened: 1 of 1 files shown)


Flowdalic:
@googlebot I signed it!

ebiggers (commented):
A couple comments:

- Any new `Makefile` parameter needs to be documented in the "Makefile Command Line Flags" section at the top of the file.

- There is already a `PAM_CONFIG_DIR` parameter which specifies where to install the Ubuntu PAM config file.  How about supporting setting that to the empty string to skip installing it, rather than adding another parameter?

Flowdalic:
Good points, that makes the change even smaller. Pushed a new revision.

Makefile
```diff
@@ -178,9 +179,11 @@ install-pam: $(PAM_MODULE)
 	install -d $(DESTDIR)$(PAM_MODULE_DIR)
 	install $(PAM_MODULE) $(DESTDIR)$(PAM_MODULE_DIR)
 
+ifdef $(PAM_CONFIG_DIR)
```
ebiggers, line 182: It needs to be `ifdef PAM_CONFIG_DIR`.  Otherwise the file won't be installed even if PAM_CONFIG_DIR is defined.

Flowdalic: Ohh, good point. Guess I was mislead by the example in the GNU Make documentation which has `ifdef $(foo)`. But the text clearly states that `$(foo)` is expanded and the result is check for being defined. Sorry about that. :/

ebiggers:
Shouldn't `make uninstall` be updated to not uninstall the file as well?  Like:

```
uninstall:                                                                       
        rm -f $(DESTDIR)$(BINDIR)/$(NAME) \                                      
              $(DESTDIR)$(PAM_INSTALL_PATH) \                                    
              $(DESTDIR)$(COMPLETION_INSTALL_DIR)/fscrypt                        
ifdef PAM_CONFIG_DIR                                                             
        rm -f $(DESTDIR)$(PAM_CONFIG_DIR)/$(NAME)                                
endif                    
```

Otherwise this looks good to me.

Flowdalic:
> Shouldn't make uninstall be updated to not uninstall the file as well?
I considered that but decided against it. Uninstall I'd using the -f switch to rm, do it will not complain if the file is not present. No need to introduce additional conditions and complexity IMHO.

ebiggers:
> > Shouldn't make uninstall be updated to not uninstall the file as well?
> 
> I considered that but decided against it. Uninstall I'd using the -f switch to rm, do it will not complain if the file is not present. No need to introduce additional conditions and complexity IMHO. Also I'd expect that uninstall always removes the whole thing.

`make uninstall` shouldn't delete any files that weren't installed by `make install`.  The problem with leaving it as-is is that if PAM_CONFIG_DIR is empty, the command would be something like:

```
rm -f /usr/local/bin/fscrypt \
      /usr/local/lib/security/pam_fscrypt.so \
      /fscrypt \
      /usr/local/share/bash-completion/completions/fscrypt
```

So if the user happens to have a file `/fscrypt` on their system, it would be deleted without warning.

Flowdalic:
This assumes that the user also unsets the variable when uninstalling. I'd expect that the user simply types "make uninstall". 

But I yes, that may not always be true. I'll push an new revision later.

Flowdalic:
Pushed

ebiggers (approved):

PR #285, opened 2021-04-21
cmd/fscrypt: use golang.org/x/term

tklauser:
The `golang.org/x/crypto/ssh/terminal` package is deprecated and merely a
wrapper around `golang.org/x/term`. Thus, use the latter directly.

### DIFF
--- a/cmd/fscrypt/format.go
+++ b/cmd/fscrypt/format.go
@@ -29,7 +29,7 @@ import (
 	"unicode/utf8"
 
 	"github.com/urfave/cli"
-	"golang.org/x/crypto/ssh/terminal"
+	"golang.org/x/term"
 
 	"github.com/google/fscrypt/util"
 )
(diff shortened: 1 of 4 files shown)


tklauser:
The build failure with Go 1.11 needs to be fixed upstream, I've sent https://golang.org/cl/312249 and will update this PR once that one is merged.

ebiggers:
> The build failure with Go 1.11 needs to be fixed upstream, I've sent https://golang.org/cl/312249 and will update this PR once that one is merged.

Great!  Another option would be to increase the minimum required Go version to build `fscrypt` to 1.12.  But it would be nice to keep 1.11 working if it's easy to do.

tklauser:
> > The build failure with Go 1.11 needs to be fixed upstream, I've sent https://golang.org/cl/312249 and will update this PR once that one is merged.
> 
> Great! Another option would be to increase the minimum required Go version to build `fscrypt` to 1.12. But it would be nice to keep 1.11 working if it's easy to do.

Thanks. The upstream fix in `golang.org/x/term` was easy enough, so I've now updated this PR to used latest master which should also build with Go 1.11.

ebiggers (approved):

ebiggers:
Merged, thanks!

PR #310, opened 2021-09-13
Readme updates

ebiggers:
* Remove note about planned commands
* Remove note about stability
* Add section about backup, restore, and recovery
* Consistently format section titles
* Add section about encrypting existing files

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,7 @@ native encryption.  See [Runtime Dependencies](#runtime-dependencies).
 	- [Enabling the PAM module on Arch Linux](#enabling-the-pam-module-on-arch-linux)
 	- [Enabling the PAM module on other Linux distros](#enabling-the-pam-module-on-other-linux-distros)
   - [Allowing `fscrypt` to check your login passphrase](#allowing-fscrypt-to-check-your-login-passphrase)
+- [Backup, restore, and recovery](#backup-restore-and-recovery)
 - [Example Usage](#example-usage)
   - [Setting up fscrypt on a directory](#setting-up-fscrypt-on-a-directory)
   - [Locking and unlocking a directory](#locking-and-unlocking-a-directory)
@@ -60,6 +61,7 @@ native encryption.  See [Runtime Dependencies](#runtime-dependencies).
   - [Getting "Package not installed" when trying to use an encrypted directory.](#getting-package-not-installed-when-trying-to-use-an-encrypted-directory)
   - [Some processes can't access unlocked encrypted files.](#some-processes-cant-access-unlocked-encrypted-files)
   - [Users can access other users' unlocked encrypted files.](#users-can-access-other-users-unlocked-encrypted-files)
+  - [Getting "Required key not available" when backing up locked encrypted files](#getting-required-key-not-available-when-backing-up-locked-encrypted-files)
   - [The reported size of encrypted symlinks is wrong.](#the-reported-size-of-encrypted-symlinks-is-wrong)
 - [Legal](#legal)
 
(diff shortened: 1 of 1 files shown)


josephlr:
@ebiggers let me know when you're done updating the documentation, and I'll review

ebiggers:
> @ebiggers let me know when you're done updating the documentation, and I'll review

I think this is everything for now.  Please take a look, thanks!

README.md
```diff
@@ -444,16 +440,108 @@ file `/etc/pam.d/fscrypt` containing:
 auth        required    pam_unix.so
 ```
 
-## Note about stability
+## Backup, restore, and recovery
+
+Encrypted files and directories can't be backed up while they are "locked", i.e.
```
josephlr, line 445: Should we give the error message that users would expect to see in this case? That way if a user gets a "Required key not available" error during backup, searching this documentation would show them the section that explains the problem?

README.md
```diff
@@ -444,16 +440,108 @@ file `/etc/pam.d/fscrypt` containing:
...
+custom passphrase will be located in `/home/.fscrypt` if you are using a
+dedicated `/home` filesystem or in `/.fscrypt` if you aren't.  If desired, you
+can back up the `fscrypt` metadata by making a copy of this directory, although
+this isn't too important since this metadata is located on the same filesystem
+as the encrypted directory(s).
+
+`pam_passphrase` (login passphrase) protectors are a bit different as they are
+always stored on the root filesystem, in `/.fscrypt`.  This ties them to the
+specific system and ensures that each user has only a single login protector.
+Therefore, encrypted directories on a non-root filesystem can't be unlocked via
```
josephlr, line 471: The "can't be unlocked via a login protector if the operating system is reinstalled or if the disk is connected to another system" is very important, should it be underlined or bolded?

README.md
```diff
@@ -444,16 +440,108 @@ file `/etc/pam.d/fscrypt` containing:
...
+passphrase in a secure location.  Then, if ever needed, you can use `fscrypt
+unlock` to unlock the directory with the recovery passphrase (by choosing the
+recovery protector instead of the login protector).
+
+Alternate approaches to supporting recovery of login passphrase-protected
+directories include manually adding your own recovery protector (using `fscrypt
+metadata add-protector-to-policy`), or backing up and restoring the `/.fscrypt`
+directory on the root filesystem (but after restore, if the UUID of the root
+filesystem changed, you will need to manually fix the UUID in any
+`.fscrypt/protectors/*.link` files on other filesystems).  The auto-generated
```
josephlr, line 490: This section's long parenthetical comment was a little confusing for me to read. Could we reword it so we don't have a whole sentence in parentheses?

README.md
```diff
@@ -444,16 +440,108 @@ file `/etc/pam.d/fscrypt` containing:
...
+The recovery passphrase is simply a `custom_passphrase` protector with a
+randomly generated high-entropy passphrase.  It is strongly recommended to
+accept the prompt to generate the recovery passphrase, then store the recovery
+passphrase in a secure location.  Then, if ever needed, you can use `fscrypt
+unlock` to unlock the directory with the recovery passphrase (by choosing the
+recovery protector instead of the login protector).
+
+Alternate approaches to supporting recovery of login passphrase-protected
+directories include manually adding your own recovery protector (using `fscrypt
+metadata add-protector-to-policy`), or backing up and restoring the `/.fscrypt`
```
josephlr, line 487: I'm wondering if we should scan all filesystems for a compatible `/.fscrypt` directory if we detect a broken link? This way restoring a `/.fscrypt` directory will work, but then we can offer to fix the issue for the user.

josephlr (approved):
Just some nits (feel free to ignore if you disagree). Thanks for updating this.

README.md
ebiggers: I've added this as a troubleshooting section which links to this section.

README.md
ebiggers: I made it bold.

README.md
ebiggers: I reworked this paragraph into a bulleted list without parentheticals.

README.md
ebiggers: Yes, I think we should do that.  Or alternatively `fscrypt` could record both the UUID and the mountpoint, and use the mountpoint as a fallback if the UUID no longer exists.  I guess the full scan would make sense because we could search for the protector ID, though.  I see you filed https://github.com/google/fscrypt/issues/311 for this already -- thanks!

Issue #115, opened 2018-12-13
Document rescue scenario with pam passpharase


RomanValov:
Say I have fscrypt setup for my home directory using pam passphrase and got my system broken or unusable or even would like to migrate my hard drive into another system. It's unclear what steps do I have to do to recover my encrypted data i.e. running from livecd environment without associated user account available.

EelcoA:
This is my current situation. I encrypted my home dir (located on a separate partition) and had to do a reinstall of Ubuntu 18.04. And now it is unclear to me what to do. Also see https://askubuntu.com/questions/1156353/failes-to-decrypt-encrypted-home-directory-after-reinstall-of-ubuntu-18-04-fscr for my issue.

Redsandro:
I wonder about this too. I have an external drive with `fscrypt` encrypted files using `pam_passphrase`. The computer was refurbished and the user no longer exists.

There are backups on AWS Glacier but it will take effort. Given that the old user was myself and I know the user password, can I access the files on the external drive? Or are the files lost because there was crucial metadata on the root filesystem?

@ebiggers could you comment on this?

---

This might be related to https://github.com/google/fscrypt/issues/125 but I ~~don't think~~ am not sure `fscrypt` will have the same hurdles as `ecryptfs`.

ebiggers:
Login protectors are stored in the `/.fscrypt` directory on the root filesystem.  If that directory has been lost, then it's impossible to decrypt any directories that were protected by the login passphrase only (i.e., not also by a custom passphrase or raw key).

Redsandro:
Thanks for clarifying @ebiggers 

> Login protectors are stored in the `/.fscrypt` directory on the __root filesystem.__

I'm hypothesizing that this is more of a problem than it is a necessity. I'm opening a new issue at #164.

So imagine we backup `/.fscrypt/`. How would we access the files on an external fscrypted drive, attached to say a completely new linux laptop?

ebiggers:
> I'm hypothesizing that this is more of a problem than it is a necessity. I'm opening a new issue at #164.

I'm afraid it's working as intended; see https://github.com/google/fscrypt/issues/164#issuecomment-554024149.  We definitely should improve the documentation though, and maybe make `fscrypt encrypt` suggest also creating a custom passphrase protector for recovery purposes.

> So imagine we backup /.fscrypt/. How would we access the files on an external fscrypted drive, attached to say a completely new linux laptop?

[...]

PR #315, opened 2021-09-14
Adjust recovery passphrase generation

ebiggers:
As per the feedback at https://github.com/google/fscrypt/issues/115
where users didn't understand that the recovery passphrase is important,
restore the original behavior where recovery passphrase generation
happens automatically without a prompt.  This applies to the case where
'fscrypt encrypt' is using a login protector on a non-root filesystem.

However, leave the --no-recovery option so that the recovery passphrase
can still be disabled if the user really wants to.  Also, clarify the
information provided about the recovery passphrase.

Update https://github.com/google/fscrypt/issues/115

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,14 +474,19 @@ via a login protector if the operating system is reinstalled or if the disk is
 connected to another system** -- even if the new system uses the same login
 passphrase for the user.
 
-Because of this, `fscrypt encrypt` will offer to generate a recovery passphrase
-when creating a login passphrase-protected directory on a non-root filesystem.
-The recovery passphrase is simply a `custom_passphrase` protector with a
-randomly generated high-entropy passphrase.  It is strongly recommended to
-accept the prompt to generate the recovery passphrase, then store the recovery
-passphrase in a secure location.  Then, if ever needed, you can use `fscrypt
-unlock` to unlock the directory with the recovery passphrase (by choosing the
-recovery protector instead of the login protector).
+Because of this, `fscrypt encrypt` will automatically generate a recovery
+passphrase when creating a login passphrase-protected directory on a non-root
+filesystem.  The recovery passphrase is simply a `custom_passphrase` protector
+with a randomly generated high-entropy passphrase.  Initially, the recovery
+passphrase is stored in a file in the encrypted directory itself; therefore, to
+use it you **must** record it in another secure location.  It is strongly
+recommended to do this.  Then, if ever needed, you can use `fscrypt unlock` to
+unlock the directory with the recovery passphrase (by choosing the recovery
+protector instead of the login protector).
+
+If you really want to disable the generation of a recovery passphrase, use the
+`--no-recovery` option.  Only do this if you really know what you are doing and
+are prepared for potential data loss.
 
 Alternative approaches to supporting recovery of login passphrase-protected
 directories include the following:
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -474,14 +474,14 @@ via a login protector if the operating system is reinstalled or if the disk is
...
-accept the prompt to generate the recovery passphrase, then store the recovery
-passphrase in a secure location.  Then, if ever needed, you can use `fscrypt
-unlock` to unlock the directory with the recovery passphrase (by choosing the
-recovery protector instead of the login protector).
+Because of this, `fscrypt encrypt` will automatically generate a recovery
+passphrase when creating a login passphrase-protected directory on a non-root
+filesystem (unless the `--no-recovery` option is given).  The recovery
+passphrase is simply a `custom_passphrase` protector with a randomly generated
+high-entropy passphrase.  It is strongly recommended to record the recovery
+passpshrase in a secure location, and not disable it.  Then, if ever needed, you
```
Redsandro, line 482: > It is strongly recommended to record the recovery passpshrase

There is a typo in `passpshrase`.

> It is strongly recommended to record the recovery passphrase in a secure location. **Period.**

Humans need to be guided and slightly forced to do the right thing indeed, but _"and not disable it"_ is a bit verbose. It is implied because it makes logical sense. The reason we are being very explicit about the importance of the recovery passphrase is to bring it to the users' attention and to convince them not to be lazy and skip this step. But once it is brought to their attention, I think it's save to assume they understand that the protector shouldn't be removed when used.

Redsandro (changes requested):
Small typo and one recommendation.

README.md
Redsandro: > I think it's save to assume they understand that the protector shouldn't be removed when used.

If you want to be really really careful, you can add a warning to the `fscrypt metadata remove-protector-from-policy` command.

> fscrypt metadata remove-protector-from-policy (...)
>
> Warning: The protector about to be removed is a passphrase_protector. You may lose your recovery passphrase. There may be no more ways to recover your data when you lose access to your root partition. Are you sure you want to continue? y/n

robszy: I would change that part:  "It is strongly recommended to record the recovery passphrase in a secure location, and not disable it"
To:
Password is stored in encrypted folder so to be able to use it you  MUST store it in pass manager or other save place

Also i would change "unless the --no-recovery option is given" to 

--no-recovery option will disable generating recovery protector an pass to it so if you are prepared for DATA LOSS you can use it if you need to bypass generating recovery for whatever reason

And place it at the end of this paragraph

ebiggers: > There is a typo in passpshrase.

Fixed.

> Humans need to be guided and slightly forced to do the right thing indeed, but "and not disable it" is a bit verbose.

Removed that part.

> If you want to be really really careful, you can add a warning to the fscrypt metadata remove-protector-from-policy command.

`fscrypt metadata` needs a lot of usability improvements in general.  I'm not planning to address this right now, as `fscrypt metadata` isn't commonly used.  And if someone is using it to remove a recovery protector, they probably got there through `fscrypt_recovery_readme.txt` which explains the recovery passphrase and recommends against removing it.

> I would change that part: "It is strongly recommended to record the recovery passphrase in a secure location, and not disable it"
> To:
> Password is stored in encrypted folder so to be able to use it you MUST store it in pass manager or other save place

I did: "Initially, the recovery passphrase is stored in a file in the encrypted directory itself; therefore, to use it you **must** record it in another secure location.  It is strongly recommended to do this."

> Also i would change "unless the --no-recovery option is given" to
>
> --no-recovery option will disable generating recovery protector an pass to it so if you are prepared for DATA LOSS you can use it if you need to bypass generating recovery for whatever reason

I did: "If you really want to disable the generation of a recovery passphrase, use the `--no-recovery` option.  Only do this if you really know what you are doing and are prepared for potential data loss."

Redsandro (approved):
I like it.

ebiggers:
@josephlr any feedback on this?

PR #346, opened 2022-02-16
Metadata validation and other security improvements

ebiggers:
This pull request makes a large number of changes (mostly related to the metadata storage) to address some reported security issues, as well as go a bit beyond the specific current issues and reduce the attack surface for the future.

For the immediate security issues, make the following changes:
* Fix `fscrypt status` and the bash completion script to handle maliciously-crafted mountpoint paths.
* Validate the size and type of all metadata files before trying to read them.
* Reject any login protectors that are owned by the wrong user, as they could be maliciously spoofed.
* Make `fscrypt setup` offer the user a choice between world-writable mode (the previous behavior) and single-user-writable mode (which for security reasons is the new default behavior, but most users should be fine with world-writable).

For additional hardening, make the following changes:
* When running as a non-root user, don't read any metadata files owned by, or in a directory owned by, another non-root user.  An opt-out flag for this is provided in `/etc/fscrypt.conf` if this particular change causes problems.
* Create all metadata files with mode 0600, rather than the default of 0644.  To make this possible, some additional changes are made to prevent files from being owned by root when they should be owned by a regular user.
* Make `pam_fscrypt` ignore system users entirely.

I've split the changes into individual commits.  They should be easier to review individually.


josephlr:
Thank you for splitting this up! Reviewing now

filesystem/filesystem.go
```diff
@@ -496,10 +502,75 @@ func (m *Mount) addMetadata(path string, md metadata.Metadata, owner *user.User)
...
+		i += count
+	}
+	return i, nil
+}
+
+// readMetadataFileSafe gets the contents of a metadata file extra-carefully,
+// considering that it could be a malicious file created to cause a
+// denial-of-service.  Specifically, the following checks are done:
+//
+// - It must be a regular file, not another type of file like a symlink or FIFO.
```
josephlr, line 694: Is there a reason a symlink can cause denial of service? I can understand it pointing to a weird file, but  we should still be able to _open_ the file. If we then don't set `O_NOFOLLOW`, we would then just check the type of the underlying file.

Also, I thought we wanted to support `.fscrypt` being a symlink? Or does `O_NOFOLLOW` only affect things if the _file_ is a symlink.

filesystem/filesystem.go
```diff
@@ -496,10 +502,75 @@ func (m *Mount) addMetadata(path string, md metadata.Metadata, owner *user.User)
...
+// Take care to avoid TOCTOU (time-of-check-time-of-use) bugs when doing these
+// tests.  Notably, we must open the file before checking the file type, as the
+// file type could change between any previous checks and the open.  When doing
+// this, O_NOFOLLOW is needed to avoid following a symlink, and O_NONBLOCK is
+// needed to avoid blocking if the file is a FIFO.  Note that this implies we
+// have to use the low-level unix.Open instead of os.OpenFile.
+//
+// This function returns the data read.
+func readMetadataFileSafe(path string) ([]byte, error) {
+	fd, err := unix.Open(path, unix.O_RDONLY|unix.O_NOFOLLOW|unix.O_NONBLOCK, 0)
```
josephlr, line 710: We can use [`os.OpenFile`](https://pkg.go.dev/os#OpenFile) here as it allows passing arbitrary flags. See [the implementation here](https://cs.opensource.google/go/go/+/refs/tags/go1.17.7:src/os/file_unix.go;l=217;drc=refs%2Ftags%2Fgo1.17.7)

Note that if we do this we do not need to wrap the path error below, as the context is added by `OpenFile` IIRC.

filesystem/filesystem.go
```diff
@@ -496,10 +502,75 @@ func (m *Mount) addMetadata(path string, md metadata.Metadata, owner *user.User)
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to open %q", path)
+	}
+	defer unix.Close(fd)
+
+	var stat unix.Stat_t
+	if err = unix.Fstat(fd, &stat); err != nil {
+		return nil, errors.Wrapf(err, "failed to stat %q", path)
```
josephlr: Here and elsewhere, if we have to use raw errors, we should return a [`os.PathError`](https://pkg.go.dev/os#PathError)

filesystem/filesystem.go
```diff
@@ -496,10 +502,75 @@ func (m *Mount) addMetadata(path string, md metadata.Metadata, owner *user.User)
...
+	}
+	if (stat.Mode & unix.S_IFMT) != unix.S_IFREG {
+		return nil, &ErrCorruptMetadata{path, errors.New("not a regular file")}
+	}
+	if stat.Size > maxMetadataFileSize {
+		return nil, &ErrCorruptMetadata{path, errors.New("metadata file size limit exceeded")}
+	}
```
josephlr, line 728: If we use `OpenFile` above, this can then just be:
```go
info, err := f.Stat()
if err != nil {
  return err
}
if !info.Mode().IsRegular() {
  // Return error
}
if info.Size() > maxMetadataFileSize {
  // Return error
}
```

filesystem/filesystem.go
```diff
@@ -496,10 +502,75 @@ func (m *Mount) addMetadata(path string, md metadata.Metadata, owner *user.User)
...
+	// Clear O_NONBLOCK, since it has served its purpose when opening the
+	// file, and the behavior of reading from a regular file with O_NONBLOCK
+	// is technically unspecified.
+	if _, err = unix.FcntlInt(uintptr(fd), unix.F_SETFL, 0); err != nil {
```
josephlr, line 732: This can also be done with normal files:
```go
unix.FcntlInt(f.Fd(), unix.F_SETFL, 0)
```

filesystem/filesystem.go
```diff
@@ -496,10 +502,75 @@ func (m *Mount) addMetadata(path string, md metadata.Metadata, owner *user.User)
...
+	// Read the file contents, making sure we hit EOF at exactly stat.Size.
+	data := make([]byte, stat.Size+1)
+	count, err := readFull(fd, data)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read from %q", path)
+	}
+	if int64(count) != stat.Size {
+		return nil, errors.Errorf("file %q was concurrently modified", path)
+	}
+	return data[:count], nil
```
josephlr: For normal files we could just use an [`io.LimitedReader`](https://pkg.go.dev/io#LimitedReader):
```go
r := &io.LimitedReader{R: f, N: maxMetadataFileSize}
data, err := io.ReadAll(r)
if err != nil {
  return err
}
if r.N == 0 {
  // Return a PathError about the file being too long
}
return data
```

Then we wouldn't even need to check the file size with Stat

filesystem/filesystem.go
```diff
@@ -581,7 +652,7 @@ func (m *Mount) AddLinkedProtector(descriptor string, dest *Mount) (bool, error)
 		}
 		return false, nil
 	}
-	if !os.IsNotExist(err) {
+	if errors.Cause(err) != unix.ENOENT {
```
josephlr, line 825: If we end up being forced to use raw errors (which we hopefully won't), we should have `readMetadataFileSafe` return an eror so that `IsNotExist` still works.

cmd/fscrypt/fscrypt_bash_completion
```diff
@@ -15,25 +15,71 @@
...
+#       not be escaped), and unescape these characters.  Exception: we don't
+#       unescape newlines, as we need to reserve newline as the separator for
+#       the words passed to compgen.  (This causes mountpoints containing
+#       newlines to not be completed correctly, which we have to tolerate.)
+#
+#   2.) We backslash-escape all shell meta-characters, and single-quote the
+#       argument passed to compgen -W.  Without either step, command injection
+#       would be possible.  Without both steps, completions would be incorrect.
+#       The list of shell meta-characters used comes from that used by the
+#       completion script for umount, which has to solve this same problem.
```
josephlr, line 59: I'm very unfamiliar with how do avoid these issues in bash. I just know it absurdly full of pitfalls and foot-guns. I know even less about how to address these issues in the context of bash completions.

Is there someone else who could review this file?

josephlr (commented):
Reviews though the first 3 commits, I'll try to have comments on the rest before end-of-week.

filesystem/filesystem.go
ebiggers: `O_NOFOLLOW` only affects the final path component.

It's true that the file could always be a weird file directly, so symlinks don't make things *that* much worse.  But symlinks can be created by non-root users pointing to anywhere, whereas creating really weird files like device nodes requires root access.  So it seems desirable to disallow symlinks in this context, given that there is no use case for them here.

filesystem/filesystem.go
```diff
@@ -496,10 +502,75 @@ func (m *Mount) addMetadata(path string, md metadata.Metadata, owner *user.User)
...
+// file type could change between any previous checks and the open.  When doing
+// this, O_NOFOLLOW is needed to avoid following a symlink, and O_NONBLOCK is
+// needed to avoid blocking if the file is a FIFO.  Note that this implies we
+// have to use the low-level unix.Open instead of os.OpenFile.
+//
+// This function returns the data read.
+func readMetadataFileSafe(path string) ([]byte, error) {
+	fd, err := unix.Open(path, unix.O_RDONLY|unix.O_NOFOLLOW|unix.O_NONBLOCK, 0)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to open %q", path)
```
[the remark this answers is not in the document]
ebiggers: This would be the first use of `%w` in `fscrypt`, whereas we're already using `errors.Wrapf()` in this way in other places.

I think this should be handled separately as part of a separate cleanup that updates the error handling everywhere.  It looks like recent Go versions added new functionality that supersedes the `"github.com/pkg/errors"` package, so we should probably stop using that package and use the built-in functionality everywhere.

filesystem/filesystem.go
ebiggers: `O_NOFOLLOW` and `O_NONBLOCK` are in the `unix` package, not `os`.  It is unclear from the documentation that `os.OpenFile()` supports raw `unix` flags.  I suppose we can use it if it works, given that if that's true then other programs will naturally be relying on this behavior, but it would be nice to see a more definitive statement somewhere that it is supported.

cmd/fscrypt/fscrypt_bash_completion
ebiggers: > Is there someone else who could review this file?

I doubt it.  Even the people who write the bash completions for much more commonly used tools mess this up.

There is one alternative, which is that we could remove the support for completing mountpoints and just use `_filedir -d`, so the completion would be to any directory instead of to mountpoints specifically.  What do you think about that?

filesystem/filesystem.go
ebiggers: Later on we need the file's uid as well, but maybe that can be gotten from `info.Sys().(*syscall.Stat_t)`

filesystem/filesystem.go
ebiggers: Changed to use `os` instead of `unix` directly.

filesystem/filesystem.go
ebiggers: Done.  (But I did `N: maxMetadataFileSize + 1` to avoid an off by one bug, and I kept `ErrCorruptMetadata` for the error.)

filesystem/filesystem.go
ebiggers: Done

filesystem/filesystem.go
ebiggers: Done

filesystem/filesystem.go
ebiggers: Done in the one place that still uses `unix` directly (`unix.FnctlInt`)

filesystem/filesystem.go
ebiggers: Done

filesystem/filesystem.go
ebiggers: Updated the comment to explain the above.

filesystem/mountpoint_test.go
```diff
@@ -179,6 +179,18 @@ func TestLoadMountWithSpecialCharacters(t *testing.T) {
 	}
 }
 
+// Tests the EscapeString() and unescapeString() functions.
+func TestStringEscaping(t *testing.T) {
+	orig := "A \t\n\\"
```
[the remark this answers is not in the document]
ebiggers: Done.  Paths can contain arbitrary bytes, not just valid UTF-8.

filesystem/mountpoint_test.go
```diff
@@ -179,6 +179,18 @@ func TestLoadMountWithSpecialCharacters(t *testing.T) {
...
+	escaped := "A\\040\\011\\012\\134"
```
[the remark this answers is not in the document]
ebiggers: Done

filesystem/filesystem.go
ebiggers: `io.ReadAll()` is only Go 1.16 and later.  Trying `ioutil.ReadAll()` instead.

pam_fscrypt/run_fscrypt.go
```diff
@@ -67,6 +68,19 @@ type PamFunc struct {
 	impl func(handle *pam.Handle, args map[string]bool) error
 }
 
+// isSystemUser checks if a user is a system user.  pam_fscrypt should never
+// need to do anything for system users since they should never have login
+// protectors.  Therefore, we detect them early to avoid wasting resources.
+func isSystemUser(user *user.User) bool {
+	uid := util.AtoiOrPanic(user.Uid)
+	if uid > 0 && uid < 1000 {
```
dirkmueller, line 76: nit: could use `SYS_UID_MAX` here

filesystem/filesystem.go
```diff
@@ -496,10 +503,60 @@ func (m *Mount) addMetadata(path string, md metadata.Metadata, owner *user.User)
...
+	info, err := file.Stat()
+	if err != nil {
+		return nil, err
+	}
+	if !info.Mode().IsRegular() {
+		return nil, &ErrCorruptMetadata{path, errors.New("not a regular file")}
+	}
+	// Clear O_NONBLOCK, since it has served its purpose when opening the
+	// file, and the behavior of reading from a regular file with O_NONBLOCK
+	// is technically unspecified.
```
mgerstner, line 721: An alternative to this approach - not really less complex, but maybe cleaner - would be to open the file using `O_PATH`, performing the `fstat()` and when all is good open the file proper via `"/proc/self/fd/<o-path-fd>"`.

filesystem/filesystem.go
josephlr: Sounds reasonable.

filesystem/filesystem.go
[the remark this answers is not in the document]
josephlr: Oh ya, I agree. After we get 0.3.3 out the door, we should just go through and update everything to idiomatic 1.16 style. Keeping as is for now seems correct.

cmd/fscrypt/fscrypt_bash_completion
josephlr: Looking at this again and thinking about it, I think the current code is fine. It's a little risky, but does seem to fix the most common pitfalls. Long term we should probably not maintain the bash completion code ourselves. I've opened https://github.com/google/fscrypt/issues/347 to track using our CLI library to do this work instead.

But for now, we should just get this fix in, and move to a different framework later.

mgerstner:
Overall the changes look good to me. Thanks for considering the additional hardenings.

filesystem/filesystem.go
ebiggers: The `procfs` magic symlinks are a little hard to wrap one's head around, and it's not clear how the case where the file itself is a symlink would get handled (`O_PATH | O_NOFOLLOW` actually "opens" the symlink itself).  So I think I'll stick with the current approach.

pam_fscrypt/run_fscrypt.go
ebiggers: That would require reading and parsing `/etc/login.defs`, right?  If that is the case, then I'm not sure we should bother with this optimization in the first place.  My hope was that the 1-999 range is standard enough that we can assume it.

josephlr: We should just have a constant defined `UID_MIN = 1000` and check that we have `uid >= UID_MIN`. If it ever becomes an issue we can fix stuff later.

It seems like 1000 is the default, and Linux distros don't really change it: https://serverfault.com/questions/804718/how-to-determine-sys-uid-min-sys-uid-max-etc-values-from-a-script

josephlr:
@ebiggers My preference would be for option (2) if we have to choose. It works better with fscrypt's other commands that modify protectors/policies. It would also only be non-atomic if the user has setup their fscrypt directory

In addition to option (2), in a follow up PR we could create a dedicated "tmp" file owned by root (with 1777 permissions) that could be used to make things atomic. In that case we would need to worry about data corruption concerns and locking, so it might not be the best choice. But we can discuss that approach later.

pam_fscrypt/run_fscrypt.go
ebiggers: Added a constant.

ebiggers:
> @ebiggers My preference would be for option (2) if we have to choose. It works better with fscrypt's other commands that modify protectors/policies. It would also only be non-atomic in the following (narrow) cases:
> 
> * the user has `/.fscrypt` setup as root-only
> * root has explicitly given a user ownership of a protector/policy file (which isn't done by default)
> * that user makes a change to one of their policies or protectors
> 
> In addition to option (2), in a follow up PR we could create a dedicated "tmp" file owned by root (with 1777 permissions) that could be used to make things atomic. In that case we would need to worry about data corruption concerns and locking, so it might not be the best choice. But we can discuss that approach later.

I've implemented option (2) as the "least bad" option.  Note that unfortunately it will be used a bit more widely than stated above.  Notably, login protectors are owned by the user, and when the user changes their passphrase, `pam_fscrypt` updates the login protector using the user's privileges.  That case will use the non-atomic overwrite.

PR #351, opened 2022-04-08
filesystem: add 'lustre' to supported filesystem types

adilger:
Add the Lustre filesystem to the list of supported types.
Lustre fully supports fscrypt API since version 2.15.0:
(ref: https://jira.whamcloud.com/browse/LU-12275
  and https://jira.whamcloud.com/browse/LU-13717).

Fixes: d0b9e2c9 ("filesystem: avoid accessing irrelevant filesystems")
Signed-off-by: Andreas Dilger <adilger@dilger.ca>

### DIFF
--- a/filesystem/filesystem.go
+++ b/filesystem/filesystem.go
@@ -393,7 +393,7 @@ func (m *Mount) isFscryptSetupAllowed() bool {
 		return true
 	}
 	switch m.FilesystemType {
-	case "ext4", "f2fs", "ubifs", "btrfs", "ceph", "xfs":
+	case "ext4", "f2fs", "ubifs", "btrfs", "ceph", "xfs", "lustre":
 		return true
 	default:
 		return false


ebiggers:
So, now that `fscrypt` has an allowlist of filesystem types, we (the maintainers of `fscrypt`) are going to be seen as responsible for what we add to it.  So by adding Lustre to the list, it could be more or less viewed as an endorsement that Lustre encryption is okay.  The problem is that Lustre is an out-of-tree filesystem, not developed in the upstream Linux kernel, and (as far as I know) the encryption support was developed without any collaboration or input from anyone else who works on Linux filesystem encryption.  You claim that Lustre fully supports the API, but does the encryption work properly as well?  Where can I find more information about this?

Note also that the README lists the supported filesystems in two places, so if Lustre is going to actually be supported, it should be added there too.

adilger:
Eric, I'd be thrilled if you would have a look over the Lustre fscrypt integration, though IMHO I think this is orthogonal to whether the "lustre" filesystem type should be "allowed" to work with the fscrypt userspace tools.  I don't think there is any implied warranty just because of this check, like there wasn't before the check was added.

In any case, the URLs in the commit message are the high-level patch tracking for the fscrypt implementation, and link to original patches in Gerrit. The code itself can be seen in the "lustre/lustre-release" project in GitHub. The main point of integration is in the client under lustre/llite. The majority of the change is in the client and transparent to the servers, but there has to be some small changes to the server to accommodate writing an encrypted block beyond EOF, and the fact that the client does not know the "block size" of the underlying storage, so it always does encryption in 4KB chunks regardless of PAGE_SIZE. 

I'm not the expert on this code, but rather Sebastien , whose email is on the commits if you want to ask questions and/or link him into the discussion here.

josephlr:
@ebiggers another thing that might be worth doing is having a blocklist of disallowed filesystems, rather than an allowlist. If I'm recalling correctly, the previous bugs were from things like `procfs`, `sysfs`, `tmpfs`, etc... which we could just disallow.

ebiggers:
> @ebiggers another thing that might be worth doing is having a blocklist of disallowed filesystems, rather than an allowlist. If I'm recalling correctly, the previous bugs were from things like `procfs`, `sysfs`, `tmpfs`, etc... which we could just disallow.

I did consider that, but there are a huge number of weird filesystem types in Linux like `cgroup`, `cgroup2`, `mqueue`, `fusectl`, `efivarfs`, etc., and people are always adding more, and userspace can add their own FUSE based filesystems that have all sorts of weird behavior.  So an allowlist seems like a better approach.

josephlr:
> > @ebiggers another thing that might be worth doing is having a blocklist of disallowed filesystems, rather than an allowlist. If I'm recalling correctly, the previous bugs were from things like `procfs`, `sysfs`, `tmpfs`, etc... which we could just disallow.
> 
> I did consider that, but there are a huge number of weird filesystem types in Linux like `cgroup`, `cgroup2`, `mqueue`, `fusectl`, `efivarfs`, etc., and people are always adding more, and userspace can add their own FUSE based filesystems that have all sorts of weird behavior. So an allowlist seems like a better approach.

What about just checking if `mount.Device` (i.e. the device name) is non-empty. At least on my system (which has a ton of weird filesystems), all the "real" file-systems have a `/dev/foo` device, while the pseudo filesystems do not.

ebiggers:
> What about just checking if `mount.Device` (i.e. the device name) is non-empty. At least on my system (which has a ton of weird filesystems), all the "real" file-systems have a `/dev/foo` device, while the pseudo filesystems do not.

That won't work for network filesystems that support encryption, such as the Lustre filesystem in question here as well as CephFS.  (Encryption support for CephFS is being worked on upstream.)

josephlr:
Then I guess we just have to maintain the allowlist. I'm fine with adding lustre. I don't think we are advertising any sort of support by just allowing fscrypt to run on that directory.

adilger:
Regarding filesystem detection, I was thinking I stead of an explicit whitelist (which IMHO will become tiring to maintain over time) whether it would be possible to call one of the existing fscrypt ioctls on the filesystem to detect if it supported fscrypt at all, and silently failing if the ioctl returned ENOTTY.

ebiggers:
> Regarding filesystem detection, I was thinking I stead of an explicit whitelist (which IMHO will become tiring to maintain over time) whether it would be possible to call one of the existing fscrypt ioctls on the filesystem to detect if it supported fscrypt at all, and silently failing if the ioctl returned ENOTTY.

That would require going back to the approach where we iterate through every mounted filesystem and access every one, which was exactly what we wanted to avoid.  Consider that there can be network based filesystems, and potentially arbitrary FUSE filesystems, that take a long time to respond.  Then `fscrypt` and `pam_fscrypt` hang whenever they are used.

ebiggers:
We could check for the presence of `/sys/fs/$FSTYP/features/encryption`, but only ext4 and f2fs support that file currently.

adilger:
It would be possible to add /sys/fs/lustre/features/encryption if it was useful to do so, though I don't think this is necessarily a "standard" interface for other filesystems. Also, it would take some time to roll this out in a release, but not impossibly so since the fscrypt support is relatively new. 

It wouldn't be terrible to have more than one mechanism to detect this.

What I find confusing (due to my ignorance of the details) is why the whitelist is needed when explicitly requesting to enable fscrypt for a specific mountpoint?

     sudo fscrypt setup /lustre/ai400x2-prod/
     Allow users other than root to create fscrypt metadata on this filesystem? (See    [https://github.com/google/fscrypt#setting-up-fscrypt-on-a-filesystem]) [y/N] y
    [ERROR] fscrypt setup: filesystem type lustre is not supported for fscrypt setup

It makes sense to recursively probe parents of the directory, if the .fscrypt file cannot be found in the requested path, but it isn't clear why all of the variety of different filesystems (/proc, /sys, etc.) would be probed in that case?

ebiggers:
There are cases where `fscrypt` or `pam_fscrypt` do operate on "all filesystems", such as when `pam_fscrypt` unlocks a user's login passphrase protected directories or when `fscrypt status` shows the global status.  Those need the whitelist, so it makes sense for other operations to use the same criteria as well --- otherwise we'd end up in a situation where some features work on a filesystem but others don't.

ebiggers:
So taking a super quick look at the Lustre code, it looks like Lustre 1.14 will lie to userspace and say that filenames are encrypted when they aren't.  Is there a way to ensure that Lustre 1.14 won't be used?

josephlr:
> We could check for the presence of `/sys/fs/$FSTYP/features/encryption`

I like this idea. We could check for this file and have a "fallback" whitelist. But this would then allow newer file systems to work with older versions of fscrypt.

ebiggers:
@adilger and @josephlr can you take a look at https://github.com/google/fscrypt/pull/355?

adilger:
Eric, the Lustre 2.14 release is not an LTS release, so it may be used for a short time but will not have any longevity. 2.15 is a designated LTS release, so any 2.14 users will migrate to that. We made it very clear in the release information that 2.14 had only data encryption and not filename encryption, so I don't think that is a major concern.

ebiggers:
Good to hear that it was made clear in the documentation and that Lustre 2.14 shouldn't be used for long.  However, if I'm reading the code correctly, Lustre 2.14 will silently ignore the filenames encryption mode passed to `FS_IOC_SET_ENCRYPTION_POLICY`, and then report back via `FS_IOC_GET_ENCRYPTION_POLICY` that it is encrypting filenames even though it's not.  So if you e.g. ran `fscrypt status DIR` you would see `Options:  padding:32 contents:AES_256_XTS filenames:AES_256_CTS policy_version:2`, which says that the filenames are being encrypted.  That seems quite a bit worse than doing it properly with a "None" encryption mode constant.

josephlr:
Closing this as #355 is merged.

sbuisson-ddn:
> Good to hear that it was made clear in the documentation and that Lustre 2.14 shouldn't be used for long. However, if I'm
> reading the code correctly, Lustre 2.14 will silently ignore the filenames encryption mode passed to
> `FS_IOC_SET_ENCRYPTION_POLICY`, and then report back via `FS_IOC_GET_ENCRYPTION_POLICY` that it is
> encrypting filenames even though it's not. So if you e.g. ran `fscrypt status DIR` you would see
> `Options: padding:32 contents:AES_256_XTS filenames:AES_256_CTS policy_version:2`, which says that the filenames
> are being encrypted. That seems quite a bit worse than doing it properly with a "None" encryption mode constant.

The reason why Lustre 2.14 reports `filenames:AES_256_CTS` while file names are actually not encrypted is precisely because fscrypt does not support any kind of "none" encryption. fscrypt does not tolerate that the file system reports a different encryption mode than the one it tried to set, so having Lustre support `fscrypt encrypt` would not have been possible in 2.14 without this.

More importantly, Lustre 2.14 does not silently ignore the filename encryption mode passed to the ioctl. When setting the policy, it prints a console warning in the form:
`forcing policy filenames_encryption_mode to null`
And when reporting the policy, it also prints a console warning:
`returning policy filenames_encryption_mode as %d, but is in fact null`

And as Andreas pointed out, this partial support of encryption in Lustre 2.14 was designed as a tech preview. Users who want to implement encryption with Lustre on production systems need to use Lustre 2.15 LTS version, that has complete encryption support.

PR #353, opened 2022-04-08
Bump up required Go version to 1.16

ebiggers:
Bump up the required Go version to 1.16 so that we can assume that Go
modules are enabled by default.  Go 1.16 is the latest end-of-life
release, so this makes it so that we support the latest end-of-life
release (1.16), the current maintainance release (1.17), the current
release (1.18), and future releases.  This the same approach we took
when we last bumped up the required Go version.

Also update the ci.yml file to test with these versions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
   build:
     strategy:
       matrix:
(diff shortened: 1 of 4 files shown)


josephlr (approved):
Looks good to me, can confirm that running `go mod tidy` doesn't change anything. So LGTM.

PR #359, opened 2022-08-12
fsync set policy ioctls

irq0:
Split policyIoctl into setPolicyIoctl and getPolicyIoctl. Add a
os.Sync() call to setPolicyIoctl.

Policy ioctls are not necessary durable on return. For example, on
ext4 (ref: fs/ext4/crypto.c: ext4_set_context) they are not. This may
lead to a filesystem containing fscrypt metadata (in .fscrypt), but
without the policy applied on an encrypted directory.

Example:
Snapshotting a mounted ext4 filesystem on Ceph RBD right after
setting the policy. While subject to timing, with high probability the
snapshot will not have the policy set. Calling fsync fixes this.

Signed-off-by: Marcel Lauhoff <marcel.lauhoff@suse.com>

### DIFF
--- a/metadata/policy.go
+++ b/metadata/policy.go
@@ -94,12 +94,25 @@ func (err *ErrNotEncrypted) Error() string {
 	return fmt.Sprintf("file or directory %q is not encrypted", err.Path)
 }
 
-func policyIoctl(file *os.File, request uintptr, arg unsafe.Pointer) error {
+func getPolicyIoctl(file *os.File, request uintptr, arg unsafe.Pointer) error {
 	_, _, errno := unix.Syscall(unix.SYS_IOCTL, file.Fd(), request, uintptr(arg))
 	if errno == 0 {
 		return nil
 	}
 	return errno
+}
+
+func setPolicy(file *os.File, arg unsafe.Pointer) error {
+	_, _, errno := unix.Syscall(unix.SYS_IOCTL, file.Fd(), unix.FS_IOC_SET_ENCRYPTION_POLICY, uintptr(arg))
+	if errno != 0 {
+		return errno
+	}
+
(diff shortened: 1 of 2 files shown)


metadata/policy.go
```diff
@@ -94,14 +94,27 @@ func (err *ErrNotEncrypted) Error() string {
...
-func policyIoctl(file *os.File, request uintptr, arg unsafe.Pointer) error {
+func getPolicyIoctl(file *os.File, request uintptr, arg unsafe.Pointer) error {
 	_, _, errno := unix.Syscall(unix.SYS_IOCTL, file.Fd(), request, uintptr(arg))
 	if errno == 0 {
 		return nil
 	}
 	return errno
 }
 
+func setPolicyIoctl(file *os.File, request uintptr, arg unsafe.Pointer) error {
```
ebiggers, line 105: `request` is always `unix.FS_IOC_SET_ENCRYPTION_POLICY` here, so please remove the unneeded parameter and rename this function to `setPolicy()`.

irq0: done

ebiggers:
The failure in "Run command-line interface tests" is unrelated and will be fixed by https://github.com/google/fscrypt/pull/362.

So it looks like you just need to sign the CLA.  (Or make sure that you've done everything for the CLA to be recognized, if your company already has it on file.)

irq0:
> So it looks like you just need to sign the CLA. (Or make sure that you've done everything for the CLA to be recognized, if your company already has it on file.)

There is already a company CLA, I'm waiting to get added.

irq0:
CLA is now signed

ebiggers (approved):

ebiggers:
Merged, thanks!

PR #363, opened 2022-08-27
clamp uint8 nCPUs variable when nCPUS >= 256

ebenali:
Panics on a dual-socket 7742 epyc box with 256 logical cores when
a int64 is cast to uint8, inducing an argon2 routine to fail upon
encountering a value of 0 (zero).

### DIFF
--- a/crypto/crypto.go
+++ b/crypto/crypto.go
@@ -39,6 +39,7 @@ import (
 	"crypto/sha512"
 	"encoding/hex"
 	"io"
+	"math"
 
 	"github.com/pkg/errors"
 	"golang.org/x/crypto/argon2"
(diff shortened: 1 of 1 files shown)


ebiggers:
Great catch!

I'm a bit concerned about systems with **more** than 256 CPUs.  Previously they would have been doing the truncation to 8 bits.  If we clamp to 255, then they'd no longer be able to unlock their directories, as the password hash function would have changed.

How about clamping `costs.Parallelism` to 255 when generating and reading the config file, in combination with using `p=1` if `uint8(costs.Parallelism) == 0`?

ebiggers:
Actually, [the Argon2 algorithm](https://datatracker.ietf.org/doc/rfc9106/) supports `p` in the range `[0, 2**24 - 1]`.  So the real bug seems to be in the Argon2 library we're using; it needs to be fixed to accept larger values of `p`.

josephlr:
I agree with @ebiggers here, we need to do this in such a way that we don't cause data-loss for older users.

To get the best behavior going forward, we could introduce a new `TruncationFixed` param to `HashingCosts`. That would indicate that `costs.Parallelism` actually holds the correct value. Otherwise, if the value is `false` (the default) we would have the previous behavior, where the parallelism cost is `uint8(costs.Parallelism)`.

josephlr:
Thanks for the work on this @ebenali, but I'm closing this in favor of #365

It seems like properly fixing this might be a little more involved.

PR #371, opened 2022-12-04
Stop using deprecated package io/ioutil

ebiggers:
Since Go 1.16 (which recently became the minimum supported Go version for this project), the package io/ioutil is deprecated in favor of equivalent functionality in the io and os packages.  staticcheck warns about this.  Address all the warnings by switching to the non-deprecated replacement functions.

### DIFF
--- a/actions/config_test.go
+++ b/actions/config_test.go
@@ -19,7 +19,6 @@
 package actions
 
 import (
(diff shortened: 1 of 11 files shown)


josephlr:
LGTM

Issue #393, opened 2023-12-04
How to unlock encrypted regular files located in an unencrypted directory?


maxnikulin:
It seems a directory may contain files encrypted by different protectors. Is it possible to find how to access file content by its path?
```
$ grep '' not-encrypted/*
grep: not-encrypted/file-a: Required key not available
grep: not-encrypted/file-b: Required key not available
not-encrypted/file-not:Not encrypted

$ fscrypt unlock crypt-a
Enter custom passphrase for protector "a": 
"crypt-a" is now unlocked and ready for use.

$ grep '' not-encrypted/*
not-encrypted/file-a:this is a
grep: not-encrypted/file-b: Required key not available
not-encrypted/file-not:Not encrypted
```
How to determine that namely the `crypt-a` directory should be unlocked to read `not-encrypted/file-a`?

Accidentally I moved a file outside of encrypted directory. It resembles a feature requested in #373. Commands to reproduce output given above:
```
$ mkdir not-encrypted crypt-a crypt-b
$ fscrypt encrypt crypt-a
The following protector sources are available:
1 - Your login passphrase (pam_passphrase)
2 - A custom passphrase (custom_passphrase)
3 - A raw 256-bit key (raw_key)
Enter the source number for the new protector [2 - custom_passphrase]: 2
Enter a name for the new protector: a
Enter custom passphrase for protector "a": 
Confirm passphrase: 
"crypt-a" is now encrypted, unlocked, and ready for use.

$ echo "this is a" >crypt-a/file-a
$ mv -iv crypt-a/file-a not-encrypted/
renamed 'crypt-a/file-a' -> 'not-encrypted/file-a'
$ fscrypt lock crypt-a
"crypt-a" is now locked.

$ fscrypt encrypt crypt-b
Should we create a new protector? [y/N] y
The following protector sources are available:
1 - Your login passphrase (pam_passphrase)
2 - A custom passphrase (custom_passphrase)
3 - A raw 256-bit key (raw_key)
Enter the source number for the new protector [2 - custom_passphrase]: 2
Enter a name for the new protector: b
Enter custom passphrase for protector "b": 
Confirm passphrase: 
"crypt-b" is now encrypted, unlocked, and ready for use.

$ echo "this is b" >crypt-b/file-b
$ mv -iv crypt-b/file-b not-encrypted/
renamed 'crypt-b/file-b' -> 'not-encrypted/file-b'
$ fscrypt lock crypt-b
"crypt-b" is now locked.

[...]

PR #396, opened 2023-12-09
Provide better error message when given a locked regular file

ebiggers:
Since opening an encrypted regular file that is locked fails with ENOKEY, getting the encryption policy of such a file is not possible. As a result, 'fscrypt status' and 'fscrypt lock' fail on such files. Provide a better error message that tries to explain what is going on.

Resolves https://github.com/google/fscrypt/issues/393


josephlr (approved):
Looks good. Thanks for fixing the message.

PR #412, opened 2024-04-30
README.md: link to RFE about systemd-homed fscrypt version support

dkg:
Including a link to the specific report can help a reader recognize when the warning about systemd-homed is no longer important.

It might also help to drive attention to the right place to improve systemd-homed.


josephlr (approved):

josephlr:
Thanks for the contribution @dkg! I updated the docs to better expand on what the problem is and link to some of our documentation and the Kernel's documentation.

Issue #411, opened 2024-04-30
Concurrency issue when new mounts are added


NymanRobin:
In the filesystem/mountpoint.go file the mountsByDevice map is recreated on each call of update UpdateMountInfo.
This causes problem, if these steps occurs!

1. Mount is assigned to policy.context
2. UpdateMountInfo is called
3. policy.apply is called

This will result in the fact that the policy.context.Mount will have the old memory address from the old mountsByDevice map while the filesystems.findMount will find the new object and then in the comparison the following error will be thrown in the policy.apply function.

(%!w(*actions.ErrDifferentFilesystem=&{0xc00157b130 0xc00208a140}).

 This currently affects the ceph-csi project and more information can be found in this issue:
 https://github.com/ceph/ceph-csi/issues/4592
 
 
 I have tested a solution were only the new mount objects are updated and that would at least work for my specific use-case, I will open a PR for this and look forward to any feedback

josephlr:
Another approach to fixing this would just be to compare the `Mount`s by value rather than by pointer.

Happy to review any PR!

NymanRobin:
Great point @josephlr!
I went ahead with the comparison as that removed the responsibility of removing unused mounts from the map and made it a tad bit easier

ebiggers:
Fixed by 068b9f8f5dec46b222470f6d3f03244ba5b65f5c

NymanRobin:
Thanks for the helping me get this change in @ebiggers and @josephlr! 
I've got one more question: when could we expect a new release of fscrypt? Do you have any plans in mind?
I'm asking because I'd like to get this fix into ceph-csi.

ebiggers:
There was a release recently, and there haven't been many changes since then, so it probably will be at least a few months before we do another release.

PR #413, opened 2024-05-02
Compare mount by value instead of reference

NymanRobin:
This has to be since the mounts are reloaded each time a mount is added. 
In case of two mounts mounting at the same time there will be a race condition for applying policy.

I added the unit test in mountpoint for my sanity while developing and some kind of documentation of the issue.

This fixes issue #411

### DIFF
--- a/actions/policy.go
+++ b/actions/policy.go
@@ -24,6 +24,7 @@ import (
 	"log"
 	"os"
 	"os/user"
+	"reflect"
 
 	"github.com/pkg/errors"
 	"google.golang.org/protobuf/proto"
@@ -452,7 +453,7 @@ func (policy *Policy) AddProtector(protector *Protector) error {
 
 	// If the protector is on a different filesystem, we need to add a link
 	// to it on the policy's filesystem.
-	if policy.Context.Mount != protector.Context.Mount {
+	if !reflect.DeepEqual(policy.Context.Mount, protector.Context.Mount) {
 		log.Printf("policy on %s\n protector on %s\n", policy.Context.Mount, protector.Context.Mount)
 		ownerIfCreating, err := getOwnerOfMetadataForProtector(protector)
 		if err != nil {
@@ -525,7 +526,7 @@ func (policy *Policy) RemoveProtector(protectorDescriptor string) error {
 func (policy *Policy) Apply(path string) error {
 	if pathMount, err := filesystem.FindMount(path); err != nil {
 		return err
-	} else if pathMount != policy.Context.Mount {
+	} else if !reflect.DeepEqual(pathMount, policy.Context.Mount) {
 		return &ErrDifferentFilesystem{policy.Context.Mount, pathMount}
 	}
 
--- a/filesystem/mountpoint_test.go
+++ b/filesystem/mountpoint_test.go
@@ -27,6 +27,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"reflect"
 	"strings"
 	"testing"
 )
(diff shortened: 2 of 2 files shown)


go.mod
josephlr, line 1: This is not an indirect dependency but a direct dependency, so it should be marked as such.

filesystem/mountpoint_test.go
```diff
@@ -544,3 +546,21 @@ func BenchmarkLoadFirst(b *testing.B) {
...
+15 0 259:3 / /home rw,relatime shared:1 - ext4 /dev/root rw,data=ordered
+`
+	beginLoadMountInfoTest()
+	loadMountInfoFromString(mountinfo)
+	fisrtMnt := mountForDevice("259:3")
+	loadMountInfoFromString(mountinfo)
+	defer endLoadMountInfoTest()
+	secondMnt := mountForDevice("259:3")
+	if !cmp.Equal(fisrtMnt, secondMnt) {
+		t.Fatal("Mount comparission does not work")
```
josephlr, line 564: I think we should use t.Errorf here

josephlr (approved):
Looks good. Thank you for submitting this

NymanRobin:
Thanks for the comments, I fixed them now!

josephlr (approved):

filesystem/mountpoint_test.go
```diff
@@ -544,3 +546,21 @@ func BenchmarkLoadFirst(b *testing.B) {
...
+	var mountinfo = `
+15 0 259:3 / /home rw,relatime shared:1 - ext4 /dev/root rw,data=ordered
+`
+	beginLoadMountInfoTest()
+	loadMountInfoFromString(mountinfo)
+	fisrtMnt := mountForDevice("259:3")
+	loadMountInfoFromString(mountinfo)
+	defer endLoadMountInfoTest()
+	secondMnt := mountForDevice("259:3")
+	if !cmp.Equal(fisrtMnt, secondMnt) {
```
ebiggers, line 563: `fisrtMnt` => `firstMnt`

filesystem/mountpoint_test.go
```diff
@@ -544,3 +546,21 @@ func BenchmarkLoadFirst(b *testing.B) {
...
+		t.Errorf("Mount comparission does not work")
```
ebiggers, line 564: `comparission` => `comparison`

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.18
 
 require (
 	github.com/client9/misspell v0.3.4
+	github.com/google/go-cmp v0.6.0
```
ebiggers, line 7: Could we use [`reflect.DeepEqual()`](https://pkg.go.dev/reflect#DeepEqual) to avoid adding another dependency?

filesystem/mountpoint_test.go
```diff
@@ -544,3 +546,21 @@ func BenchmarkLoadFirst(b *testing.B) {
 		}
 	}
 }
+
+// Test mount comparison by values instead of by reference,
+// as the map mountsByDevice gets recreated on each load.
+// This ensures that concurrent mounts work properly.
+func TestMountComparission(t *testing.T) {
```
ebiggers, line 553: `Comparisson` => `Comparison`

filesystem/mountpoint_test.go
```diff
@@ -544,3 +546,21 @@ func BenchmarkLoadFirst(b *testing.B) {
...
+	var mountinfo = `
+15 0 259:3 / /home rw,relatime shared:1 - ext4 /dev/root rw,data=ordered
+`
+	beginLoadMountInfoTest()
+	loadMountInfoFromString(mountinfo)
+	fisrtMnt := mountForDevice("259:3")
+	loadMountInfoFromString(mountinfo)
+	defer endLoadMountInfoTest()
```
ebiggers, line 561: `defer endLoadMountInfoTest()` should go right after `beginLoadMountInfoTest()`

actions/policy.go
```diff
@@ -525,7 +526,7 @@ func (policy *Policy) RemoveProtector(protectorDescriptor string) error {
 func (policy *Policy) Apply(path string) error {
```
ebiggers, line 526: What about the comparison in `AddProtector()`?

```
        // If the protector is on a different filesystem, we need to add a link  
        // to it on the policy's filesystem.                                     
        if policy.Context.Mount != protector.Context.Mount {                     
```

go.mod
NymanRobin: Sure, this go-cmp was already an indirect dependency so did not think it would have been to bad add, but for this simple case I think reflect works fine.

```
$ go mod graph | grep "cmp"
google.golang.org/protobuf@v1.33.0 github.com/google/go-cmp@v0.5.5
```

actions/policy.go
NymanRobin: Good point, thanks. Did this not occur in my testing as both policy and protector is set from the same reference in the ceph-csi. But I guess if someone would do a lookup instead between the calls this would fail, so I will change this also

NymanRobin:
Thanks for the comments @ebiggers, I fixed up the stuff hope it looks good now :)

PR #417, opened 2024-12-03
Document stdin behaviour for getting raw key

betelgeuse:
Document the logic described in `makeRawKey`.


ebiggers:
Merged, thanks!

Issue #422, opened 2025-02-14
non-constant format string diagnostic in Go 1.24


mikelolasagasti:
Go 1.24 now reports a diagnostic for calls of the form `fmt.Printf(s)`, where `s` is a non-constant format string, with no other arguments.

In the first two Go 1.24 RC releases, this check was enforced regardless of the `go.mod` version. Since RC3 and the final release, the check is only enforced if `go.mod` specifies go 1.24 or higher, so it’s not an immediate issue but will become one when the required Go version is eventually increased.

~~~
go vet -printf ./...
# github.com/google/fscrypt/cmd/fscrypt
# [github.com/google/fscrypt/cmd/fscrypt]
cmd/fscrypt/commands.go:417:62: non-constant format string in call to github.com/pkg/errors.Wrapf
cmd/fscrypt/commands.go:524:61: non-constant format string in call to github.com/pkg/errors.Wrapf
~~~

Ref: https://go.dev/doc/go1.24#vet

PR #423, opened 2025-02-14
fix: use constant format string in errors.Wrapf

mikelolasagasti:
Go 1.24 introduced stricter requirements for format strings in errors.Wrapf, requiring the first argument to be a constant string. This change updates Wrapf calls to use a constant format string (`"path: %s"`) instead of passing the `path` variable directly.

Refs: https://go.dev/doc/go1.24#vet

Resolves https://github.com/google/fscrypt/issues/422

### DIFF
--- a/cmd/fscrypt/commands.go
+++ b/cmd/fscrypt/commands.go
@@ -414,7 +414,7 @@ func unlockAction(c *cli.Context) error {
 	if policy.IsProvisionedByTargetUser() {
 		log.Printf("policy %s is already provisioned by %v",
 			policy.Descriptor(), ctx.TargetUser.Username)
-		return newExitError(c, errors.Wrapf(ErrDirAlreadyUnlocked, path))
+		return newExitError(c, errors.Wrapf(ErrDirAlreadyUnlocked, "path: %s", path))
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/fscrypt/commands.go
```diff
@@ -414,7 +414,7 @@ func unlockAction(c *cli.Context) error {
 	if policy.IsProvisionedByTargetUser() {
 		log.Printf("policy %s is already provisioned by %v",
 			policy.Descriptor(), ctx.TargetUser.Username)
-		return newExitError(c, errors.Wrapf(ErrDirAlreadyUnlocked, path))
+		return newExitError(c, errors.Wrapf(ErrDirAlreadyUnlocked, "path: %s", path))
```
ebiggers, line 417: Shouldn't these just be replaced with `errors.Wrap()`?

Otherwise the printed message looks weird:

```
[ERROR] fscrypt unlock: path: /test/dir: this file or directory is already unlocked
```

For comparison, currently it is:
```
[ERROR] fscrypt unlock: /test/dir: this file or directory is already unlocked
```

ebiggers: https://github.com/google/fscrypt/pull/424 fixes it correctly using errors.Wrap().

PR #449, opened 2026-04-14
cmd/fscrypt: mention --unlock-with in protector error hint

### LINKED ISSUE #439: Error message tells you to use the wrong argument
While testing how I could automate unlocking some files in a script, I tried to pass in a password from stdin, and got an error saying I have to tell it which protector that password goes with:

```
$ echo abc | sudo fscrypt unlock test1/ --quiet
[ERROR] fscrypt unlock: multiple protectors available

Use --protector=MOUNTPOINT:ID to specify a protector.
```

The problem is that --protector is not the right switch to use.  
```
$ echo abc | sudo fscrypt unlock --protector=/mnt/ceph:3ebd7d4bd4de24bf test1/ --quiet
fscrypt unlock: flag provided but not defined: -protector
```

It looks like the correct switch is --unlock-with:
`$ echo abc | sudo fscrypt unlock --unlock-with=/mnt/ceph:3ebd7d4bd4de24bf test1/ --quiet
`
^ that one works

alliasgher:
Fixes #439. The error hint for multiple protectors only mentioned `--protector`, but for the `unlock` command the correct flag is `--unlock-with`. Now both flags are shown.

### DIFF
--- a/cmd/fscrypt/errors.go
+++ b/cmd/fscrypt/errors.go
@@ -272,7 +272,8 @@ func getErrorSuggestions(err error) string {
 		return fmt.Sprintf("If desired, use %s to automatically run destructive operations.",
 			shortDisplay(forceFlag))
 	case ErrSpecifyProtector:
-		return fmt.Sprintf("Use %s to specify a protector.", shortDisplay(protectorFlag))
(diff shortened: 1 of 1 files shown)


ebiggers (approved):

ebiggers:
Merged, thanks.

PR #455, opened 2026-07-11
Don't retry passwords if `fscrypt unlock` doesn't have a tty

### LINKED ISSUE #429: `fscrypt unlock </dev/null` hangs indefinitely
Steps to reproduce:
1. Create a directory managed by fscrypt
2. Run `fscrypt unlock DIR </dev/null`

I expect that this exits with an error, because it's never going to get a password (ideally it would say that it's unable to read stdin). Instead, it loops forever until canceled:
```
Enter login passphrase for jyn: Incorrect Passphrase
Enter login passphrase for jyn: Incorrect Passphrase
Enter login passphrase for jyn: Incorrect Passphrase
Enter login passphrase for jyn: Incorrect Passphrase
...
```

This bit me originally because I put `fscrypt unlock` in `/etc/ssh/sshrc`, so it completely locked me out of my machine until I was able to get physical access to it.

ebiggers:
Fixes https://github.com/google/fscrypt/issues/429

### DIFF
--- a/cli-tests/t_unlock.out
+++ b/cli-tests/t_unlock.out
@@ -62,6 +62,30 @@ Protected with 1 protector:
 PROTECTOR         LINKED  DESCRIPTION
 desc2  No      custom protector "prot"
 
+# Try to unlock with no stdin
+[ERROR] fscrypt unlock: incorrect key provided
+"MNT/dir" is encrypted with fscrypt.
+
+Policy:   desc1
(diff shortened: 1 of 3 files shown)


josephlr (approved):
LGTM. Are we ok with prompting for a retry if the password is wrong and the user passed `--quiet`?

ebiggers:
> LGTM. Are we ok with prompting for a retry if the password is wrong and the user passed `--quiet`?

Changed to make `--quiet` continue to disable retrying.

PR #456, opened 2026-08-09
docs: document udev dependency for /dev/disk/by-uuid links

### LINKED ISSUE #55: Document exact specifics of udev dependancy
With #54, the directory `/dev/disk/by-uuid` and its contents are now necessary to establish cross filesystem links. This means use of these links now depends on a specific _default_ udev rule, namely `60-persistent-storage.rules`. The source can be found in the [systemd tree](https://github.com/systemd/systemd/blob/master/rules/60-persistent-storage.rules), [eudev tree](https://github.com/gentoo/eudev/blob/master/rules/60-persistent-storage.rules), and [the old udev tree](https://github.com/lu-zero/udev/blob/master/rules/60-persistent-storage.rules).

The documentation (probably part of #10) should explain this an other dependencies in a single section. It should also explain when these dependencies are necessary (build-time, run-time, link creation time, etc..).

amarkdotdev:
## Summary
- Document that cross-filesystem UUID links depend on `/dev/disk/by-uuid` and the standard `60-persistent-storage.rules` udev rule.
- Clarify this is a runtime / link-creation dependency (not build-time), including path fallback behavior.

Fixes #55

## Notes
- Contributor will sign the Google CLA if required for this repository.

## Test plan
- [ ] Review the new Runtime dependencies subsection in README.md for accuracy

### DIFF
--- a/README.md
+++ b/README.md
@@ -337,8 +337,8 @@ run a minimal or custom udev setup without that rule, ensure equivalent rules
 still create `/dev/disk/by-uuid/*` symlinks before relying on UUID-based
 cross-filesystem links.  Example upstream sources:
 
-* [systemd `60-persistent-storage.rules`](https://github.com/systemd/systemd/blob/master/rules.d/60-persistent-storage.rules)
-* [eudev `60-persistent-storage.rules`](https://github.com/gentoo/eudev/blob/master/rules/60-persistent-storage.rules)
+* [systemd `60-persistent-storage.rules`](https://github.com/systemd/systemd/blob/main/rules.d/60-persistent-storage.rules.in)
+* [eudev `60-persistent-storage.rules`](https://github.com/eudev-project/eudev/blob/master/rules/60-persistent-storage.rules)
 
 ## Configuration file
 
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ For the release notes, see the [NEWS file](NEWS.md).
 - [Features](#features)
 - [Building and installing](#building-and-installing)
 - [Runtime dependencies](#runtime-dependencies)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -313,6 +314,32 @@ If you configure `fscrypt` to use non-default features, other kernel
...
+* **Runtime (following a link):** when resolving an existing link that contains a
+  `UUID=` entry, `fscrypt` again uses `/dev/disk/by-uuid` (with a path fallback if
+  the UUID cannot be resolved).
+
+Most Linux distributions ship `60-persistent-storage.rules` by default.  If you
+run a minimal or custom udev setup without that rule, ensure equivalent rules
+still create `/dev/disk/by-uuid/*` symlinks before relying on UUID-based
+cross-filesystem links.  Example upstream sources:
+
+* [systemd `60-persistent-storage.rules`](https://github.com/systemd/systemd/blob/master/rules.d/60-persistent-storage.rules)
```
ebiggers: This link doesn't work.

README.md
```diff
@@ -313,6 +314,32 @@ If you configure `fscrypt` to use non-default features, other kernel
...
+* [eudev `60-persistent-storage.rules`](https://github.com/gentoo/eudev/blob/master/rules/60-persistent-storage.rules)
```
ebiggers: This one does work, but it redirects to a new location.  Best to link directly to the new location.

README.md
amarkdotdev: swapped it for a working link

README.md
amarkdotdev: pointed at the dest instead of the redirect

ebiggers:
Merged, thanks.

